PR #82, opened 2022-10-24
[GEN-154] Update the case selection code to check postmortem seq cases

xindiguo:
- add review column for potential postmortem cases
- add seed to irr and sdv randomization

### DIFF
--- a/scripts/case_selection/perform_case_selection.R
+++ b/scripts/case_selection/perform_case_selection.R
@@ -1,7 +1,7 @@
 # Description: Perform BPC case selection by first constructing an eligibility matrix
 #   and then returning all records that fit the eligibility criteria.  Also, sample
 #   the requested number of SDV and IRR cases.  
-# Author: Haley Hunter-Zinck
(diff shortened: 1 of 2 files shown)


scripts/case_selection/perform_case_selection.R
```diff
@@ -296,12 +304,14 @@ create_selection_matrix <- function(eligible_cohort, n_prod, n_pressure, n_sdv,
   
   # randomly disperse additional SDV cases among non-pressure
   col_sdv <- rep("", n_eligible)
+  set.seed(site_seed)
   idx_sdv <- sample((n_pressure+1):n_prod, n_sdv)
```
thomasyu888, line 308: Why do you need to set.seed so many times?

thomasyu888 (approved):
Lgtm, just a small question

scripts/case_selection/perform_case_selection.R
xindiguo: You need to set seed every time before running sample to make the order reproducible.

PR #94, opened 2023-05-27
remove support for synapse config

thomasyu888:
This simplifies the codebase by quite a bit, because we are using environmental variable to login to the client


rxu17 (approved):
LGTM!

PR #96, opened 2023-06-08
[GEN-736] Add debug mode and production parameter for bpc

rxu17:
**Purpose:** This allows the genie bpc pipeline to be run on nextflow in production or development mode with the addition of the following:

- `debug` mode in all processes in main.nf
- `production` parameter for use in all processes in main.nf (with the exception of `quacUploadReportError` which doesn't upload to synapse in general) which for right now, it indicates whether we upload/update anything in Synapse (prod) or not (develop)

This also adds nextflow configuration and development instructions to the README and fixes a bug with $comment in the `main.nf`

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,15 +210,15 @@ You can visit [parameters](https://github.com/Sage-Bionetworks/genie-bpc-pipelin
 
 1. For an ec2 instance with Linux and docker, see here for installing Java 11: [How do I install a software package from the Extras Library on an EC2 instance running Amazon Linux 2?](https://aws.amazon.com/premiumsupport/knowledge-center/ec2-install-extras-library-software/)
 
-1. Install nextflow by following instructions here: [Get started — Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#get-started)
+1. Install nextflow by following instructions here: [Get started — Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#get-started). Update your `PATH` variable to include the directory where your nextflow executable is installed at.
 
-- Update PATH to include directory where your nextflow executable is
+1. Make sure to set any nextflow secrets using the Nextflow Cli: [Secrets — Nextflow](https://www.nextflow.io/docs/latest/secrets.html#command-line). You will need to set a `SYNAPSE_AUTH_TOKEN` secret for running the nextflow genie repo by doing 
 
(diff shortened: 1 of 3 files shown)


rxu17:
Wasn't sure if we also wanted to split it up into modules for this PR?

thomasyu888:
Let's not split it up into modules this PR, but this looks good! Feel free to test!

main.nf
```diff
@@ -123,10 +154,18 @@ process updateDataTable {
...
+   if (params.production) {
+      """
+      cd /root/scripts/
+      python update_data_table.py -p /root/scripts/config.json -m $comment primary
+      """
+   }
+   else {
+      """
+      cd /root/scripts/
+      python update_data_table.py -p /root/scripts/config.json -m $comment -d True primary
```
thomasyu888: @rxu17 can you speak to why `-d True` is necessary here?  Is it because of the "primary" str

rxu17: I was doing some testing while making that change, but for some reason in the end I need to add quotes and change this line to:
`python update_data_table.py -p /root/scripts/config.json -m "$comment" primary -d`

The original issue was that it was not interpreting $comment as a multi-word string but as separate arguments so `table_type` is set to `public` which isn't allowed. Not sure if this will successfully run itself all the way through but it stopped throwing the error, and the command line call looks okay now based on the debug statements.

thomasyu888: Thanks for the explanation!  May want to add a comment about that - string arguments can be a headache.  This PR is looking great.

README.md
```diff
@@ -170,4 +174,64 @@ To conduct case selection for a new site, modify the `config.yaml` file. Add in
...
+For GENIE BPC, here are the specification recommendations when launching an EC2 instance:
+
+- EC2 Product: Linux with Docker
+- EC2 Instance Type: t3.2xlarge
+- Disk size: ~100 GB
+
+### Nextflow Dependencies
+
+- [Java 8 or later](https://www.java.com/en/download/)
+- [Nextflow 22.10.x or earlier](https://www.nextflow.io/docs/latest/getstarted.html#get-started)
```
thomasyu888, line 194: Why earlier?  This will probably change to support later versions of nextflow when we upgrade to DSL2

thomasyu888 (approved):
🦖 LGTM!  Thanks for the great work and documentation!  This also validates the other ticket, so i'll merge that in first.

README.md
rxu17: Yup, my plan is to just update this once we change to DSL2

thomasyu888:
Not sure why this closed, but I will re-open

Issue #29, opened 2022-03-25
Create separate Dockerfiles for synthetic scripts


hhunterzinck:
create separate Dockerfiles for python and r and use in separate processes in nextflow workflow

PR #97, opened 2023-06-22
Add nextflow parameter validation

rxu17:
**Purpose:** This adds parameter validation from groovy scripts from`nf-core` + some custom validation code, and adds help prompts for running the nextflow pipeline. This also adds labels and headings to`main.nf` and removes most of the custom validation code we used previously given that we use `nextflow_schema.json` now


thomasyu888 (approved):
🔥 Nice!  This looks great

rxu17:
Ran the nextflow pipeline locally and it ran all the way through minus the `run_quac_table_report` step which stays idle and is getting investigated here [#29](https://github.com/Sage-Bionetworks/genie-bpc-quac/pull/29)

PR #98, opened 2023-06-22
[GEN-602] Update to dsl2

rxu17:
**Purpose:** This updates the pipeline to use DSL2 from Nextflow (adds modules, updates `main.nf` to use modules, etc.)

Depends on  #97


README.md
```diff
@@ -4,7 +4,7 @@
 
 Clone this repository and navigate to the directory:
 ```
-git clone git@github.com:Sage-Bionetworks/genie-bpc-pipeline.git
+git clone https://github.com/Sage-Bionetworks/genie-bpc-pipeline.git
```
thomasyu888, line 7: nit: One thing to note here.  the `git@` is not technically wrong, some people just use `ssh` keys to connect with github.

thomasyu888 (approved):
🔥 this is awesome - thanks for looking into this!

PR #99, opened 2023-06-29
GEN-757 Add CRC phase 1 update to config file

xindiguo:
Copied oncotree code info from phase 1
Used May-2018 to Dec-2022 seq date


Chelsea-Na (approved):
Looks good!

PR #104, opened 2023-10-12
[GEN-888] use chatgpt to convert R to Python and modify script a tiny bit

thomasyu888:
Instead of fixing the R depencies, it's a lot faster to just convert from R to Python and run the script locally.


Chelsea-Na (approved):
Looks good

PR #113, opened 2023-11-15
[GEN-970] update birth year redaction code to allow nan value

xindiguo:
TypeError occurred due to emtpy value in the column. Original expected value type was integer

Solution:
Modify the code to allow NaN value and convert float to int before updating the table

### DIFF
--- a/scripts/table_updates/update_data_table.py
+++ b/scripts/table_updates/update_data_table.py
@@ -112,13 +112,12 @@ def _to_redact_birth_year(df_col_birth_year, df_col_dt_compare, df_col_vital_sta
     Returns:
         tuple: pandas.core.indexes: to redact index in Dataframe
     """
-    df_col_birth_year = pandas.to_numeric(df_col_birth_year, downcast='integer',errors='coerce')
     if len(df_col_dt_compare)==0:
         df_col_dt_compare = datetime.date.today()
     else:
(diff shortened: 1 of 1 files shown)


scripts/table_updates/update_data_table.py
```diff
@@ -112,13 +112,12 @@ def _to_redact_birth_year(df_col_birth_year, df_col_dt_compare, df_col_vital_sta
...
         tuple: pandas.core.indexes: to redact index in Dataframe
     """
-    df_col_birth_year = pandas.to_numeric(df_col_birth_year, downcast='integer',errors='coerce')
     if len(df_col_dt_compare)==0:
         df_col_dt_compare = datetime.date.today()
     else:
         df_col_dt_compare = df_col_dt_compare.fillna(datetime.datetime.now().strftime("%Y-%m-%d"))  
         df_col_dt_compare = df_col_dt_compare.apply(lambda x: datetime.datetime.strptime(str(x),"%Y-%m-%d").date())
-    dates_diff = df_col_dt_compare - df_col_birth_year.apply(lambda x: datetime.date(x, 1, 1))
+    dates_diff = df_col_dt_compare - df_col_birth_year.apply(lambda x: datetime.date(datetime.date.today().year, 1, 1) if math.isnan(x) else datetime.date(int(x), 1, 1))
```
thomasyu888, line 120: @xindiguo can you explain why you're taking the current year when it is null?

thomasyu888 (approved):
Going to pre-approve, but just had a question.

PR #115, opened 2024-01-12
look up folder children with links and download using Synapse File Link

xindiguo:
Issues:

- Look up entity children was limited to File entity only
- Was not able to download Link type of entity

Solution:
- Expand the look up to be all types of entities
- Add a helper function to download Link type

### DIFF
--- a/scripts/release/update_retraction_table.py
+++ b/scripts/release/update_retraction_table.py
@@ -86,11 +86,27 @@ def get_file_id_by_name(syn, folder_id, file_name):
         folder_id (String): Synapse Folder ID
         file_name (String): File Name
     """
-    for f in syn.getChildren(folder_id, includeTypes=['file']):
+    for f in syn.getChildren(folder_id):
         f_name = f['name']
         f_id = f['id']
         if f_name == file_name:
             return(f_id)
+        
+def download_sample_file(syn, file_id):
+    """Download sample file by the file Synapse ID
+
+    Args:
+        syn: Synapse Object
+        file_id (String): Synapse file ID
+    """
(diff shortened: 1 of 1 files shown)


Chelsea-Na (approved):
Looks good. Though unsure what the caveats are (if any) to expanding to use all types of entities. Maybe using the followLink parameter in this case would be better.

scripts/release/update_retraction_table.py
```diff
@@ -86,11 +86,27 @@ def get_file_id_by_name(syn, folder_id, file_name):
...
+        syn: Synapse Object
+        file_id (String): Synapse file ID
+    """
+    file_entity = syn.get(file_id)
+    if file_entity.concreteType == 'org.sagebionetworks.repo.model.Link':
+        linked_id = file_entity.linksTo['targetId']
+        linked_version = file_entity.linksTo['targetVersionNumber']
+        return(pandas.read_csv(syn.get(linked_id,version=linked_version).path, sep='\t', header=None, usecols=[0]))
+    else:
+        return(pandas.read_csv(syn.get(file_id).path, sep='\t', header=None, usecols=[0]))
```
thomasyu888, line 108:
```suggestion
    entity = syn.get(file_entity, followLink=True)
    return(pandas.read_csv(entity.path, sep='\t', header=None, usecols=[0]))
```

thomasyu888:
```suggestion
    file_entity = syn.get(file_id, followLink=True)
    return(pandas.read_csv(file_entity.path, sep='\t', header=None, usecols=[0]))
```

thomasyu888 (approved):

PR #116, opened 2024-01-13
[GEN-1068, GEN-467] Convert case selection to python

thomasyu888:
**Problem**
R + synapser has been more difficult to maintain, and we have more Python expertise.

**Solution**

1. Leveraged github copilot + codeium to convert R to Python
2. Clean up the code 
3. Put common functions into utils.py
4. Create `validation_argparse_input`

We should review on a call


scripts/case_selection/update_case_count_table.py
```diff
@@ -0,0 +1,191 @@
...
+                    cohort=cohort,
+                    phase=phase,
+                    site=site,
+                )
+                n_target = get_production(config, phase, cohort, site)
+                n_adjust = get_adjusted(config, phase, cohort, site)
+                n_pressure = get_pressure(config, phase, cohort, site)
+                n_sdv = get_sdv(config, phase, cohort, site)
+                n_irr = get_irr(config, phase, cohort, site)
+                new_row = pd.DataFrame(
```
thomasyu888, line 112: Need to validate the content

PR #119, opened 2024-01-30
Update remove_patients_from_clinical_files.R

xindiguo:
- Use the new BPC Release Retraction table instead of the temporary retraction table
- update the query column to patient_id


Chelsea-Na (approved):
LGTM!

PR #123, opened 2024-02-22
GEN-1179 remove custom yesno mapping

xindiguo:
Issue:

Variable field type update (yesno -> dropdown) in later version of data dictionary resulted in mapping to NA 

Solution:
Add yesno variables to the GRS and remove the custom mapping in the code. Use GRS as the universal key to map


Chelsea-Na (approved):
Looks good.

PR #124, opened 2024-03-08
[GEN-1193] Update cohort variable in the derived variable files

xindiguo:
Issue:
To proceed with the release pipeline with two phases of the same cohort. We need to update the cohort variable with the following rules in the derived variable files
- NSCLC2 -> NSCLC
- CRC2 -> CRC

Solution:
Script to pull down the rdata, update cohort, upload as csv files and rdata file


thomasyu888 (changes requested):
This can be closed

PR #125, opened 2024-03-14
[GEN-1198] Update config.yaml

Chelsea-Na:
Updating to remove PROV from the RENAL cohort and setting irr to 0 for the RENAL, OVARIAN, MELANOMA, ESOPHAGO cohorts.


xindiguo (approved):
Look great!

PR #129, opened 2024-04-15
[GEN-1260] Update config.yaml

Chelsea-Na:
Applying updates for the OVARIAN cohort:
- Updated the case counts
- Set WAKE to 0 cases
- Set pressure cases = 5: VHIO, PROV, UCSF
- Set IRR to 0 for all sites


xindiguo (approved):
Looks good!

PR #132, opened 2024-06-23
hotfix due to more than one table

xindiguo:
cancer_diagnosis now has more than one table


Chelsea-Na (approved):
Looks good! Though I wonder why the field `redcap_data_access_group` gets removed from the split Synapse tables.

PR #133, opened 2024-06-24
update the syntax to avoid Error 9 in GEN-1367

xindiguo:


Chelsea-Na (approved):
I tested this update on a partial production run and it worked. Approved. 🔥

PR #135, opened 2024-07-03
[GEN-1404] Update config.yaml

Chelsea-Na:
Updating VICC CRC Phase 1 target case counts to n=226 from n=225.


thomasyu888:
LGTM! sorry for the delay in review!

PR #136, opened 2024-07-11
[GEN-1415] Update code to use pandas.concat

thomasyu888:
Problem
The code was using an older supported syntax from pandas

Solution
Update to use `pandas.concat` which is the new way to appending dataframes together


Chelsea-Na (approved):
LGTM

PR #138, opened 2024-07-13
[GEN-1424] Make sure 'None' gets read in as a string

thomasyu888:
Problem:
None was being read in as a NaN value with the newer version of pandas when I did an upgrade

Solution
Need to remove 'None' as a NaN value in pandas

### DIFF
--- a/scripts/table_updates/update_data_table.py
+++ b/scripts/table_updates/update_data_table.py
@@ -289,15 +289,18 @@ def main():
     with open(project_config) as config_file:
         cohort_info = json.load(config_file)
         logger.info("Read cohort information successful.")
-        config_file.close()
     
     # get master table
+    # This is the internal tables with non redacted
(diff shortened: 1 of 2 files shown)


scripts/table_updates/update_data_table.py
```diff
@@ -289,15 +289,18 @@ def main():
     with open(project_config) as config_file:
         cohort_info = json.load(config_file)
         logger.info("Read cohort information successful.")
-        config_file.close()
```
Chelsea-Na, line 292: Might wanna close this

Chelsea-Na (approved):
LGTM - though you should add back in line 292.

scripts/table_updates/update_data_table.py
thomasyu888: The context manager automatically closes this which is the benefit of using the

```
with ... as:
```

instead of:
```
file = open(....)
file.close()
```

PR #139, opened 2024-07-25
[GEN-1448] Fix bug that causes blank Curation and QA and Patient Characteristics tables

rxu17:
**Purpose:** See [JIRA Ticket](https://sagebionetworks.jira.com/browse/GEN-1448) for more background info but this adds back "" as a default string to NA value when reading csvs. The `get_data` function wasn't reading in blank values as NAs so later on in the code when we filter using `isnull()`, we wouldn't get any data showing up.

**Testing:**
Tested on pipeline with `""` added and `" "` removed, and got back non-blank data for the above tables

### DIFF
--- a/scripts/table_updates/utilities.py
+++ b/scripts/table_updates/utilities.py
@@ -69,7 +69,7 @@ def get_data(syn, label_data_id, cohort):
     Returns:
         Dataframe: label data
     """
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #140, opened 2024-07-29
[GEN-1456] downgrade synapseclient

rxu17:
**Purpose:** See [JIRA ticket](https://sagebionetworks.jira.com/browse/GEN-1456) for the background info. `synapseclient` is downgraded to `2.7.2` thus downgrading `pandas` to <2.0 so there is no clash in the code. 

**Testing:** Ran update_data_tables.py locally and synapse tables have expected results


danlu1 (approved):
LGTM

PR #141, opened 2024-08-01
[GEN-1458] Refactor custom_fix function to pull from main genie release

rxu17:
**Purpose:** Due to an issue with the main genie prod clinical sample database data, and because `custom_fix` pulls from that table, we've refactored `custom_fix` to pull from the clinical sample file from a main genie consortium release.

**Changes:**
- `get_main_genie_clinical_sample_file` - this is a new function to get the clinical sample data for us. [See changes](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1458-fix-custom-fix/scripts/table_updates/update_data_table.py#L33)
- `custom_fix` (renamed to `custom_fix_for_cancer_panel_test_table`) - refactored to use `get_main_genie_clinical_sample_file` function instead of reading from the clinical sample db. [See changes](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1458-fix-custom-fix/scripts/table_updates/update_data_table.py#L344-L348)

**Testing:**
- Unit tests
- Running test run

### DIFF
--- a/scripts/table_updates/config.json
+++ b/scripts/table_updates/config.json
@@ -18,5 +18,8 @@
         "NSCLC2": "syn51318736",
         "CRC2": "syn52943210",
         "RENAL": "syn59474249"
-    }
+    },
+    "main_genie_release_version": "16.6-consortium",
+    "main_genie_data_release_files": "syn16804261",
+    "main_genie_sample_mapping_table": "syn7434273"
 }
--- a/scripts/table_updates/update_data_table.py
+++ b/scripts/table_updates/update_data_table.py
@@ -24,68 +24,115 @@ import numpy
 
 from utilities import *
 
-TABLE_INFO = {"primary": ('syn23285911',"table_type='data'"),
-              "irr": ('syn21446696',"table_type='data' and double_curated is true")}
+TABLE_INFO = {
+    "primary": ("syn23285911", "table_type='data'"),
+    "irr": ("syn21446696", "table_type='data' and double_curated is true"),
+}
+
+
+def get_main_genie_clinical_sample_file(
+    syn: synapseclient.Synapse, release: str, release_files_table_synid: str
+) -> pandas.DataFrame:
+    """This retrieves the main genie clinical sample file from consortium release
+
+    Args:
+        release (str): release version to pull from for main genie
+        release_files_table_synid (str): synapse id of the data relese files table
+        from main genie
+
+    Returns:
+        pandas.DataFrame: the read in clinical file as dataframe
+    """
+    release_files = syn.tableQuery(
+        f"SELECT * FROM {release_files_table_synid}"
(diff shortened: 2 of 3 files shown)


scripts/table_updates/update_data_table.py
```diff
@@ -24,307 +24,438 @@
...
 
-TABLE_INFO = {"primary": ('syn23285911',"table_type='data'"),
-              "irr": ('syn21446696',"table_type='data' and double_curated is true")}
+TABLE_INFO = {
+    "primary": ("syn23285911", "table_type='data'"),
+    "irr": ("syn21446696", "table_type='data' and double_curated is true"),
+}
+
+
+def get_main_genie_clinical_sample_file(
```
thomasyu888, line 33: Nit: this is great for this function, but I think we have this particular "get main genie file from release" function used across all GENIE operations.  In the future, when we re-design this whole thing, it might be worthwhile to add all GENIE helper functions in one package.

scripts/table_updates/update_data_table.py
```diff
@@ -24,307 +24,438 @@
...
+    """
+    release_files = syn.tableQuery(
+        f"SELECT * FROM {release_files_table_synid}"
+    ).asDataFrame()
+    #assert_frame_equal(release_files, release_files)
+    clinical_link_synid = release_files[
+        (release_files["release"] == release)
+        & (release_files["name"] == "data_clinical_sample.txt")
+    ]["fileSynId"].values[0]
+    clinical_link_ent = syn.get(clinical_link_synid)
```
thomasyu888: Use `followLink=True` parameter to download the linked file.

rxu17: This way you don't need to specify the `version` parameter as well?

thomasyu888: That's right, it will download the correct version

thomasyu888 (approved):
🔥 LGTM!  I'm going to pre-approve this - but I do know it's lacking tests.  In the future, we will need to figure out the CI/CD for this to auto test BPC code.

scripts/table_updates/config.json
```diff
@@ -18,5 +18,8 @@
         "NSCLC2": "syn51318736",
         "CRC2": "syn52943210",
         "RENAL": "syn59474249"
-    }
+    },
+    "main_genie_release_version": "16.6-consortium",
+    "main_genie_data_release_files": "syn16804261",
```
thomasyu888, line 23: Nit: Does it make sense to use the database to synapse id mapping table?  (Maybe in the future - but not for this iteration, if we are going to re-design the whole process)

rxu17: I think so for future iterations at least for retrieving the specific tables yes. I just wanted to limit dependencies for this one.

PR #142, opened 2024-08-13
[GEN-1470] Update config.yaml(s)

Chelsea-Na:
Only updating the config files to refer to the new sites for each cohort. 

I think what we need to do next is update all the scripts to accept the new sites/cohorts/phases as user input. Ideally we align them all so that they use the config file as possible options. Also there are some choices across separate scripts that accept different inputs for the same idea (e.g. "2" vs "phase 2"). This should also be aligned.

### DIFF
--- a/scripts/masking/config.yaml
+++ b/scripts/masking/config.yaml
@@ -86,6 +86,33 @@ cohort:
       - MSK
       - UHN
       - UCSF
+  OVARIAN:
+    sites:
+      - DFCI
+      - MSK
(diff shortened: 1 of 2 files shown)


thomasyu888 (commented):
Thanks, let's not merge this yet because I don't want this merge to accidentally break the code.  I'm unfortunately just not certain about the changes these config changes cause for the code.

thomasyu888 (approved):

PR #143, opened 2024-08-13
Add one config

thomasyu888:
This is to create a central config for certain things so that it reduces the amount of duplication of code and increases consistency.  There are still some config files that should probably not be added into this.


rxu17:
Closing this as won't do.

PR #146, opened 2024-08-20
[GEN-1484] Fix mismatch between patient list in case selection and consortium release

thomasyu888:
**Problem**:
The issue is we use the synapse database tables to create the case selection list, this will cause problems if any of these patients or samples are filtered out of the consortium releases which is what is being used to generate the export list

**Solution**
Add print statements to indicate process and also include logic to filter and include patients that are in consortium release


scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -75,7 +75,7 @@ if (!is.element(site, site_option)) {
 }
 
 # setup ----------------------------
-
+print("get clinical data")
 # clinical data
 # always use the most recent consortium release - Feb 2022
```
thomasyu888:
```suggestion
# HACK: always use the most recent consortium release at the time of execution of this code
```

Chelsea-Na (approved):
LGTM 🔥

PR #149, opened 2024-08-23
[GEN-1489] Improve validation logs and remove patients from all output

thomasyu888:
**Problem**
All patients were being included in the export file even though they should be filtered out

**Solution**
Filter the patients out and improve the logs

```
[1] "map data for each instrument"
[1] "recode"
[1] "instrument cancer panel test"
[1] "validate output"
[1] "Patient(s) removed from the BPC export file:"...
[1] "Sample(s) removed from the BPC export file:"...
[1] "N missing patients:  2"
[1] "N missing samples:  6"
[1] "N unique patients in export file  756"
[1] "N unique samples in export file 796"
[1] "N Unique patients in case selection 758"
[1] "N Unique samples in case selection 802"
```


Chelsea-Na (approved):
LGTM!

PR #150, opened 2024-08-30
Create dd-compare.py

Chelsea-Na:
Adding a utility script that compares two PRISSMM data dictionaries

### DIFF
--- a/scripts/dd/dd-compare.py
+++ b/scripts/dd/dd-compare.py
@@ -44,6 +44,14 @@ Example cleaned output:
 image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
 """
 def clean_multiline_content(file_content):
+    """
+    Function to clean up the multiline content while preserving specific rows
+    Example multiline input:
+    image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located?
+    Cancer Site 6", ...
+    Example cleaned output:
+    image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
+    """
     cleaned_lines = []
     temp_line = ''
     inside_quotes = False
--- a/scripts/dd/dd-compare.py
+++ b/scripts/dd/dd-compare.py
@@ -33,16 +33,6 @@ args = parser.parse_args()
 entity1 = syn.get(args.synid1)
 entity2 = syn.get(args.synid2)
 
-"""
-Function to clean up the multiline content while preserving specific rows
-
-Example multiline input:
-image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located?
-Cancer Site 6", ...
-
-Example cleaned output:
-image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
-"""
 def clean_multiline_content(file_content):
     """
     Function to clean up the multiline content while preserving specific rows
--- a/scripts/dd/dd-compare.py
+++ b/scripts/dd/dd-compare.py
@@ -1,6 +1,12 @@
-# Purpose: Takes in two PRISSMM data dictionaries and print out a simple comparison report
-# Author: Sage Bionetworks
-# Date: 29AUG2024
+"""
+Purpose: Takes in two PRISSMM data dictionaries and print out a simple comparison report
+Author: Sage Bionetworks
+Date: 29AUG2024
+
+Environments:
+pandas==3.9
+synapseclient==4.1.1
+"""
 
 import pandas as pd
 import synapseclient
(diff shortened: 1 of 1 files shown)


scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,160 @@
...
+syn.login()
+
+# Set the default PRISSMM data dictionaries
+
+default1 = syn.get('syn52903784').path #v4.0.1
+default2 = syn.get('syn61600728').path #v4.1.4
+
+# Initialize arg parser
+parser = argparse.ArgumentParser(description="Compare two data dictionaries.")
+parser.add_argument('--df1', type=str, default=default1, help='File path to the older data dictionary CSV')
```
thomasyu888: You may want to rename these as synapse IDs since that's the input. 

--dd1_synid, --dd2_synid

thomasyu888: If you set the default here to be synapse ids, then you can just do a syn.get at the bottom instead of what you're doing now. Then the argpaser doesnt rely on a syn.get

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,160 @@
...
+parser.add_argument('--df2', type=str, default=default2, help='File path to the newer data dictionary CSV')
+
+# Parse the arguments
+args = parser.parse_args()
+
+entity1 = syn.get(args.df1)
+entity2 = syn.get(args.df2)
+
+# Function to clean up the multiline content while preserving specific rows
+def clean_multiline_content(file_content):
```
thomasyu888, line 36: You can add docstrings by doing 

"""
Your comment here
"""

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,160 @@
...
+parser.add_argument('--df1', type=str, default=default1, help='File path to the older data dictionary CSV')
+parser.add_argument('--df2', type=str, default=default2, help='File path to the newer data dictionary CSV')
+
+# Parse the arguments
+args = parser.parse_args()
+
+entity1 = syn.get(args.df1)
+entity2 = syn.get(args.df2)
+
+# Function to clean up the multiline content while preserving specific rows
```
thomasyu888: It might be helpful to provide an example of what said row looks like

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,160 @@
...
+# Function to read and clean a CSV file, then load into a DataFrame
+def read_and_clean_csv(file_path):
+    with open(file_path, 'r') as file:
+        content = file.readlines()
+
+    # Clean the content
+    cleaned_content = clean_multiline_content(content)
+
+    # Create a temporary CSV file in memory
+    temp_csv = StringIO("\n".join(cleaned_content))
```
thomasyu888, line 78: Nice usage of StringIO here. Im surprised pandas doesn't have the functionality to read these things nicely.

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,160 @@
+# Purpose: Takes in two PRISSMM data dictionaries and print out a simple comparison report
```
thomasyu888: Can you add docstrings up top for synapse client version and pandas version?

scripts/dd/dd-compare.py
Chelsea-Na: Updated to use --synid1 and --synid2

Chelsea-Na: Updated to only use syn.get at the bottom.

scripts/dd/dd-compare.py
Chelsea-Na: Provided an example within a docstring.

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,171 @@
...
+Function to clean up the multiline content while preserving specific rows
+
+Example multiline input:
+image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located?
+Cancer Site 6", ...
+
+Example cleaned output:
+image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
+"""
+def clean_multiline_content(file_content):
```
thomasyu888, line 36:
```suggestion
def clean_multiline_content(file_content):
    """
    Function to clean up the multiline content while preserving specific rows
    Example multiline input:
    image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located?
    Cancer Site 6", ...
    Example cleaned output:
    image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
    """
```

scripts/dd/dd-compare.py
```diff
@@ -0,0 +1,179 @@
...
+"""
+Function to clean up the multiline content while preserving specific rows
+
+Example multiline input:
+image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located?
+Cancer Site 6", ...
+
+Example cleaned output:
+image_casite6,prissmm_imaging,,dropdown,"Where is the cancer located? Cancer Site 6", ...
+"""
```
thomasyu888:
```suggestion
```

thomasyu888 (approved):
LGTM!

PR #151, opened 2024-09-03
[GEN-1481] Update case selection scripts to pull from a main genie consortium release

danlu1:
Problem:

Currently the case selection script pulls from the main GENIE clinical patient and sample databases. We need to link this to a certain main GENIE clinical files for a given release version for the sake of release tracking.

Solution:

1. Add the get_main_genie_clinical_id function to pull synapse id for data_clinical.txt file for a given release
2. Add release parameter to related scripts allowing specify main GENIE release version when implementing the case selection.


README.md
```diff
@@ -54,16 +54,22 @@ Options:
 	-s SITE, --site=SITE
 		BPC site
 
-	-u, --save_synapse
-		Save output to Synapse
-
+	-mp MAIN_PATIENT, --main_patient=MAIN_PATIENT
+		Main GENIE clinical patient file
+
+	-ms MAIN_SAMPLE, --main_sample=MAIN_SAMPLE
```
thomasyu888: This is probably a good stopgap solution, but it would be ideal if the parameter was

```
--release 17.2-consortium
```

Then there's code that looks for the clinical patient and sample files

thomasyu888: Since you've done a lot of work already here, lets hold off on that refactor and make it as a future improvement?

danlu1: @thomasyu888 I'm using synapse id since Chelsea prefers synapse id as she mentioned in last week's coordination meeting. I'm happy to tweak this if needed.

thomasyu888: @Chelsea-Na can you add your thoughts here?

Chelsea-Na: I apologize for any confusion. Both approaches have their pros and cons. While the BPC code typically uses a synID as input for other scripts, it's often more convenient to specify the release name directly. Both methods are valid.

However, the optimal solution would be to have a single input parameter to refer to the release, as shown in Tom's example: either "--release 17.2-consortium" or "--synid syn62173551". This approach is preferred because I can't foresee a scenario where we'd need to specify separate release versions for the patient and sample files when running case selection.

danlu1: Gotcha, thanks for clarifying. I think we can go ahead with a format like `--release 17.2-consortium` so it's easy to see what consortium release version the code is using. @thomasyu888 It won't take too long to modify this but I'll wait for your go-ahead.

thomasyu888: Thanks @danlu1 ! Feel free to use this fileview: https://www.synapse.org/Synapse:syn17019650/tables/ to find the release folders

subworkflows/case_selection_workflow.nf
```diff
@@ -10,6 +10,7 @@ params.center = "DFCI"
 params.production = false
 params.bpc_input = "syn53294194"
 params.bpc_output = "syn62147862"
+params.release = ""
```
thomasyu888:
```suggestion
params.release = "17.2-consortium"
```

scripts/case_selection/workflow_case_selection.R
```diff
@@ -76,8 +79,27 @@ prov_exec_selection <- "https://github.com/Sage-Bionetworks/genie-bpc-pipeline/t
 prov_exec_add <- prov_exec_selection
 prov_exec_report <- "https://github.com/Sage-Bionetworks/genie-bpc-pipeline/tree/develop/scripts/case_selection/perform_case_selection.Rmd"
 
+#' Get Main GENIE clinical file using release version name
+#'
+#' @param release Release version name for a GENIE consortium release
+#'                
+#' @return A named list of Main GENIE clinical file Synapse ID.
+get_main_genie_clinical_id <- function(release){
```
rxu17: Could this be in a common functions script? And then all three scripts that use it can source from it so we don't have to modify in three different places?

I don't want to introduce a new R library but `here` helps with the filepath management in R. I think you can just do

```
source(here(scripts/common_functions.R))
```
it would also be easier to write tests for it? R is very finicky, so I think it may be worthwhile to write a test for this since you functionalized it to make sure we're getting our expected output.

thomasyu888 (approved):
🔥 LGTM! thanks for doing this.  Happy to merge after @Chelsea-Na confirms all the data checks on her end.

scripts/case_selection/workflow_case_selection.R
```diff
@@ -76,8 +79,27 @@ prov_exec_selection <- "https://github.com/Sage-Bionetworks/genie-bpc-pipeline/t
...
+  query <- glue("SELECT id FROM syn17019650 WHERE name = '{release}'")
+  release_folder_id <- as.character(unlist(as.data.frame(synTableQuery(query, includeRowIdAndRowVersion = F))))
+  release_files <- as.list(synGetChildren(release_folder_id, includeTypes = list("link")))
+  for (release_file in release_files){
+    if (release_file$name == "data_clinical.txt"){
+      return(release_file$id)
+    }
+  }
```
rxu17: nit: Add `return(NULL)` outside for loop if function has no match or throw error

scripts/case_selection/perform_case_selection.Rmd
```diff
@@ -7,6 +7,7 @@ params:
   phase: x
```
rxu17, line 7: Is this the only place in the case section process that we can see the versions of things that were ran?

rxu17 (approved):
LGTM! Had a few questions/comments

scripts/case_selection/workflow_case_selection.R
thomasyu888: Since it'll be quite the lift to have CI/CD for the tests and mocking in R, what do you recommend as a quick testing strategy?

rxu17: I was going to suggest filtering by version on the `data_clinical.txt` file directly but then I saw that the file's version comments are all `database`. :')

Ahh the quickest I can think of is have a test (basically integration) that runs this function for a sample release say `17.1-consortium_release` and then check it against the actual synapse id it should output.

danlu1: This could be a common function. I didn't put it in there simply because I don't want to touch that file. lol. I can test the here package. If it fails, I will add it as a new ticket.

scripts/case_selection/perform_case_selection.Rmd
danlu1: Yup, as my understanding.

scripts/case_selection/workflow_case_selection.R
danlu1: > Ahh the quickest I can think of is have a test (basically integration) that runs this function for a sample release say 17.1-consortium_release and then check it against the actual synapse id it should output.

I tested it on my local when developing the function and it did output as expected.

thomasyu888: I think rixing was referring to using something like testthat R package and doing an integration test. 

Might be good practice if we are splitting off the function so the expectation is written out.

Although if adding dependencies at this point adds additional issues to the package.... That's something we should discuss.

thomasyu888: In this particular scenario, I would be ok to be a bit lenient, but there are some upcoming work that we should absolutely add tests for as it pertains to mapping of values and business logic we need to ensure is right.

Leveraging Chelsea for heavy output QA and QC will be required

danlu1: Got it. How about we discuss about adding test to code thoroughly next week? I found there are several test scripts both in tests folder and scripts folder, such as this: [test_dd_to_nonphi_dd.R](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/develop/scripts/dd/test_dd_to_nonphi_dd.R)

thomasyu888: Sounds good - one issue is that none of the tests are actually run on PR, so let's come up with a testing strategy that allows us to ensure high quality data delivery but also have confidence in the codebase and still allow us to meet our October delivery.

I have my own thoughts but I'll wait to hear both  your thoughts.

rxu17: Down to discuss overall testing strategy. Yeah I'm not sure if we can rely on those tests because they were updated a couple years ago. 

Right now I think we can only do ~two things for tests (just putting down some thoughts):
- Write tests (unit/integration) for new functions we add/functions we add new code to unless they are super long which then maybe we rely on the below method instead. Maybe like Tom suggested before, have something to just run those tests right before the rest of the code runs instead of integrating CI/CD
- Like Tom mentioned, integrate code to compare before and after (data quality testing)

I still think it would be good to add the test(s) (`testthat`) for that function in a separate test script before we merge the PR even if we can't automate the testing of it so we don't forget about it later with all of the BPC things coming up

scripts/case_selection/workflow_case_selection.R
danlu1: I added a check at a earlier step and it will error out if release version is invalid.

rxu17: In the same function?

danlu1: Yup. Like this:
```
get_main_genie_clinical_id <- function(release){
  query <- glue("SELECT id FROM syn17019650 WHERE name = '{release}'")
  release_folder_id <- as.character(unlist(as.data.frame(synTableQuery(query, includeRowIdAndRowVersion = F))))
  if (length(release_folder_id) == 0) {
  stop(glue("The release version is invalid."))
}
  release_files <- as.list(synGetChildren(release_folder_id, includeTypes = list("link")))
  for (release_file in release_files){
    if (release_file$name == "data_clinical.txt"){
      return(release_file$id)
    }
  }
}
```

scripts/case_selection/workflow_case_selection.R
danlu1: I can add a test script under the case_selection folder then. Let's discuss an overall strategy next week.

scripts/case_selection/workflow_case_selection.R
rxu17: Oh gotcha, I guess I meant having a return statement just for good practice since we're relying on default behavior of R functions where if `if (release_file$name == "data_clinical.txt")` is never true, it will return `NULL`

danlu1: hmmm I thought data_clinical.txt will always be available if we do a consortium release.

rxu17: That's why this is just a `nit`. Yeah, it technically should always be available but with data/code you never know haha

scripts/case_selection/perform_case_selection.Rmd
thomasyu888: There's also provenance, which should link back to the different versions.

scripts/case_selection/workflow_case_selection.R
danlu1: Added

danlu1:
@rxu17 I added the test scripts. Please take a look.

scripts/case_selection/test_case_selection.R
rxu17, line 1: Nit: Rename script to `test_shared_fxns.R` and put in `scripts/case_selection/tests`

rxu17 (approved):
New edits/tests look great!

scripts/case_selection/test_case_selection.R
danlu1: Okay. On it.

PR #152, opened 2024-09-11
[GEN-1476] use detailed columns to extract tier1a code

danlu1:
Problem:

Currently numerical values of the NAACCR codes that are mapped to CBIO_LABEL instead of granular DESCRIPTIONS for target ETHNICITY, PRIMARY_RACE, SECONDARY_RACE, TERTIARY_RACE, SEX, SAMPLE_TYPE columns. The numeric code and CBIO_LABEL have one-many relation which could cause errors in remapping step. 

Solution:

Use DESCRIPTIONS column in the mapping tables to extract numeric code.

### DIFF
--- a/scripts/case_selection/export_bpc_selected_cases.R
+++ b/scripts/case_selection/export_bpc_selected_cases.R
@@ -42,12 +42,12 @@ if (is.null(opt$input) || is.null(opt$phase) || is.null(opt$cohort) || is.null(o
   stop("Usage: Rscript export_bpc_selected_cases.R -h")
 }
 
-in_file <- 'syn62828306'
-out_folder <- 'syn62828556'
-phase <- 'phase 1'
-cohort <- 'NSCLC'
-site <- 'DFCI'
-release <- '17.2-consortium'
+in_file <- opt$input
+out_folder <- opt$output
+phase <- opt$phase
+cohort <- opt$cohort
+site <- opt$site
+release <- opt$release
 
 # check user input -----------------
 
--- a/scripts/case_selection/export_bpc_selected_cases.R
+++ b/scripts/case_selection/export_bpc_selected_cases.R
@@ -120,33 +120,7 @@ existing_patients = selected_cases[selected_cases %in% clinical$patient_id]
 samples_per_patient <- clinical$sample_id[clinical$patient_id %in% selected_cases]
 
 print("map data for each instrument")
-#'  mapping data for instrument - patient_characteristics
-#' 
-#' @param clinical A data frame of released clinical data for selected cases
-#' @param existing_patients A data frame of available patient after case selection
-#' @return A data frame with mapped code
-remap_patient_characteristics <- function(clinical, existing_patients, ethnicity_mapping, race_mapping, sex_mapping){
-  
-  patient_df <- data.frame("record_id" = existing_patients)
-  patient_df$redcap_repeat_instrument <- rep("")
-  patient_df$redcap_repeat_instance <- rep("")
-  
-  patient_df$genie_patient_id <- patient_df$record_id
-  patient_df$birth_year <- clinical$birth_year[match(patient_df$genie_patient_id, clinical$patient_id)]
-  patient_df$naaccr_ethnicity_code <- clinical$ethnicity_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
-  patient_df$naaccr_race_code_primary <- clinical$primary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
-  patient_df$naaccr_race_code_secondary <- clinical$secondary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
-  patient_df$naaccr_race_code_tertiary <- clinical$tertiary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
-  patient_df$naaccr_sex_code <- clinical$sex_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
-  
-  # mapping to code
-  patient_df$naaccr_ethnicity_code <- ethnicity_mapping$CODE[match(patient_df$naaccr_ethnicity_code, ethnicity_mapping$DESCRIPTION)]
-  patient_df$naaccr_race_code_primary <- race_mapping$CODE[match(patient_df$naaccr_race_code_primary, race_mapping$DESCRIPTION)]
-  patient_df$naaccr_race_code_secondary <- race_mapping$CODE[match(patient_df$naaccr_race_code_secondary, race_mapping$DESCRIPTION)]
-  patient_df$naaccr_race_code_tertiary <- race_mapping$CODE[match(patient_df$naaccr_race_code_tertiary, race_mapping$DESCRIPTION)]
-  patient_df$naaccr_sex_code <- sex_mapping$CODE[match(patient_df$naaccr_sex_code,sex_mapping$DESCRIPTION)]
-}
-
+# instrument - patient_characteristics
 patient_output <- remap_patient_characteristics(clinical, existing_patients, ethnicity_mapping, race_mapping, sex_mapping)
 
 print("recode")
--- a/scripts/case_selection/export_bpc_selected_cases.R
+++ b/scripts/case_selection/export_bpc_selected_cases.R
@@ -42,12 +42,12 @@ if (is.null(opt$input) || is.null(opt$phase) || is.null(opt$cohort) || is.null(o
   stop("Usage: Rscript export_bpc_selected_cases.R -h")
 }
 
-in_file <- opt$input
-out_folder <- opt$output
-phase <- opt$phase
-cohort <- opt$cohort
-site <- opt$site
-release <- opt$release
+in_file <- 'syn62828306'
+out_folder <- 'syn62828556'
+phase <- 'phase 1'
+cohort <- 'NSCLC'
+site <- 'DFCI'
+release <- '17.2-consortium'
 
 # check user input -----------------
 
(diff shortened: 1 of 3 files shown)


scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -165,6 +165,7 @@ sample_info_list <- lapply(samples_per_patient,function(x){
     temp_df$cpt_genie_sample_id = x[i]
     temp_df$cpt_oncotree_code = clinical$oncotree_code[clinical$sample_id == x[i]]
     temp_df$cpt_sample_type = clinical$sample_type_detailed[clinical$sample_id == x[i]]
+    temp_df$cpt_sample_type = sample_type_mapping$CODE[match(temp_df$cpt_sample_type, sample_type_mapping$DESCRIPTION)]
```
thomasyu888: Can you speak to why this was changed here?

danlu1: Since as the Jira ticket, we would like to output numeric value for this column:



Input: Main GENIE [Value] in Consortium Release | Output: BPC [Numerical Code] in Export File
-- | --
PRIMARY_RACE_DETAILED | naaccr_race_code_primary
SECONDARY_RACE_DETAILED | naaccr_race_code_secondary
TERTIARY_RACE_DETAILED | naaccr_race_code_tertiary
SEX_DETAILED | naaccr_sex_code
ETHNICITY_DETAILED | naaccr_ethnicity_code
SAMPLE_TYPE_DETAILED | cpt_sample_type

thomasyu888: Thanks, but I thought the previous line covered that: 
```
clinical$sample_type_detailed[clinical$sample_id == x[i]]
```

thomasyu888: Ohhh interesting, before we weren't providing these as codes.  @Chelsea-Na , I wonder... if SAMPLE_TYPE_DETAILED is meant to be provided as a string value instead

danlu1: The `sample_type_detailed` renders info in DESCRIPTION column in this [table](https://www.synapse.org/Synapse:syn7434273/tables/).

scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -128,18 +128,18 @@ patient_output$redcap_repeat_instance <- rep("")
...
-patient_output$naaccr_race_code_tertiary <- race_mapping$CODE[match(patient_output$naaccr_race_code_tertiary, race_mapping$CBIO_LABEL)]
-patient_output$naaccr_sex_code <- sex_mapping$CODE[match(patient_output$naaccr_sex_code,sex_mapping$CBIO_LABEL)]
+patient_output$naaccr_ethnicity_code <- ethnicity_mapping$CODE[match(patient_output$naaccr_ethnicity_code, ethnicity_mapping$DESCRIPTION)]
+patient_output$naaccr_race_code_primary <- race_mapping$CODE[match(patient_output$naaccr_race_code_primary, race_mapping$DESCRIPTION)]
+patient_output$naaccr_race_code_secondary <- race_mapping$CODE[match(patient_output$naaccr_race_code_secondary, race_mapping$DESCRIPTION)]
+patient_output$naaccr_race_code_tertiary <- race_mapping$CODE[match(patient_output$naaccr_race_code_tertiary, race_mapping$DESCRIPTION)]
+patient_output$naaccr_sex_code <- sex_mapping$CODE[match(patient_output$naaccr_sex_code,sex_mapping$DESCRIPTION)]
 print("recode")
 # recode
 # cannotReleaseHIPAA = NA
```
thomasyu888, line 133: Expand code below. @Chelsea-Na is this a concern that we are recoding?

```
# cannotReleaseHIPAA = NA
patient_output$birth_year[which(patient_output$birth_year == "cannotReleaseHIPAA")] <- NA
# -1 Not collected = 9 Unknown
patient_output$naaccr_ethnicity_code[which(patient_output$naaccr_ethnicity_code == -1)] <- 9
# -1 Not collected = 99 Unknown
patient_output$naaccr_race_code_primary[which(patient_output$naaccr_race_code_primary == -1)] <- 99
# -1 Not collected = 88 according to NAACCR
patient_output$naaccr_race_code_secondary[which(patient_output$naaccr_race_code_secondary == -1)] <- 88
patient_output$naaccr_race_code_tertiary[which(patient_output$naaccr_race_code_tertiary == -1)] <- 88
```

scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -128,18 +128,18 @@ patient_output$redcap_repeat_instance <- rep("")
 
 patient_output$genie_patient_id <- patient_output$record_id
 patient_output$birth_year <- clinical$birth_year[match(patient_output$genie_patient_id, clinical$patient_id)]
-patient_output$naaccr_ethnicity_code <- clinical$ethnicity[match(patient_output$genie_patient_id, clinical$patient_id)]
-patient_output$naaccr_race_code_primary <- clinical$primary_race[match(patient_output$genie_patient_id, clinical$patient_id)]
-patient_output$naaccr_race_code_secondary <- clinical$secondary_race[match(patient_output$genie_patient_id, clinical$patient_id)]
-patient_output$naaccr_race_code_tertiary <- clinical$tertiary_race[match(patient_output$genie_patient_id, clinical$patient_id)]
-patient_output$naaccr_sex_code <- clinical$sex[match(patient_output$genie_patient_id, clinical$patient_id)]
+patient_output$naaccr_ethnicity_code <- clinical$ethnicity_detailed[match(patient_output$genie_patient_id, clinical$patient_id)]
```
thomasyu888: @danlu1 thoughts on adding some integration tests to make sure we are doing these mappings correctly?

scripts/case_selection/export_bpc_selected_cases.R
Chelsea-Na: @thomasyu888 This recoding is not a concern, this is expected. However, we did recently update to NAACCR version 24 in the main GENIE registry which has `99` for [Race 1 Data Item #160 ](https://apps.naaccr.org/data-dictionary/data-dictionary/version=24/data-item-view/item-number=160/) mapping to "Unknown by patient" instead of just "patient". I would update the comment on line 149 to reflect that.

scripts/case_selection/export_bpc_selected_cases.R
Chelsea-Na: Hmmm yeah, I'm not sure why we were providing `SAMPLE_TYPE` as the Description value instead of the numerical code. I would recommend we just stick to how we have always done provided it (using the Description value).

scripts/case_selection/export_bpc_selected_cases.R
danlu1: @Chelsea-Na I extended the comments to reflect what are in the mapping tables.

scripts/case_selection/export_bpc_selected_cases.R
danlu1: I'm thinking about put the mapping lines into a remap_naaccr_code function then test if the function generates expected result. What do you think?

thomasyu888: I like that idea! Thanks!

scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -120,35 +120,44 @@ existing_patients = selected_cases[selected_cases %in% clinical$patient_id]
...
-patient_output$naaccr_race_code_primary <- race_mapping$CODE[match(patient_output$naaccr_race_code_primary, race_mapping$CBIO_LABEL)]
-patient_output$naaccr_race_code_secondary <- race_mapping$CODE[match(patient_output$naaccr_race_code_secondary, race_mapping$CBIO_LABEL)]
-patient_output$naaccr_race_code_tertiary <- race_mapping$CODE[match(patient_output$naaccr_race_code_tertiary, race_mapping$CBIO_LABEL)]
-patient_output$naaccr_sex_code <- sex_mapping$CODE[match(patient_output$naaccr_sex_code,sex_mapping$CBIO_LABEL)]
+#'  mapping data for instrument - patient_characteristics
+#' 
+#' @param clinical A data frame of released clinical data for selected cases
+#' @param existing_patients A data frame of available patient after case selection
+#' @return A data frame with mapped code
+remap_patient_characteristics <- function(clinical, existing_patients, ethnicity_mapping, race_mapping, sex_mapping){
```
thomasyu888: I know it's a WIP, but you may want to think about moving this to `shared_fxn.R` so you can more easily write a `thatthat` function.  Testing this will be a bit tedious, as it would require some fake data.  We want to make sure the matches are happening correctly.

if we use existing GENIE data, we can potentially test with a checksum after writing out the transformed file

scripts/case_selection/export_bpc_selected_cases.R
```diff
@@ -120,35 +120,44 @@ existing_patients = selected_cases[selected_cases %in% clinical$patient_id]
...
+  patient_df$genie_patient_id <- patient_df$record_id
+  patient_df$birth_year <- clinical$birth_year[match(patient_df$genie_patient_id, clinical$patient_id)]
+  patient_df$naaccr_ethnicity_code <- clinical$ethnicity_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
+  patient_df$naaccr_race_code_primary <- clinical$primary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
+  patient_df$naaccr_race_code_secondary <- clinical$secondary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
+  patient_df$naaccr_race_code_tertiary <- clinical$tertiary_race_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
+  patient_df$naaccr_sex_code <- clinical$sex_detailed[match(patient_df$genie_patient_id, clinical$patient_id)]
+  
+  # mapping to code
+  patient_df$naaccr_ethnicity_code <- ethnicity_mapping$CODE[match(patient_df$naaccr_ethnicity_code, ethnicity_mapping$DESCRIPTION)]
```
thomasyu888: Do we want to throw an error if there isn't a match?

scripts/case_selection/export_bpc_selected_cases.R
danlu1: Ya, I'm thinking about moving this to shared_fxn.R since I ran into issue when sourcing the script in test script. I've generated a test case, but source the `remap_patient_characteristics` function seems problematic cuz it stopped in line 41 if no argument is given.

scripts/case_selection/export_bpc_selected_cases.R
danlu1: I can add check here.

danlu1: I had the second thought. This might not be an issue due to line `existing_patients = selected_cases[selected_cases %in% clinical$patient_id]`. Let me test it out.

thomasyu888: Hmmm, I am concerned about strings not mapping to DESCRIPTION, but maybe that isn't an issue

danlu1: That shouldn't be an issue since the detailed columns are actually generated from DESCRIPTION in the table.

thomasyu888: As long as we don't update the description column in the mapping tables.

danlu1: I see your point. Ya, This could be an issue if we update DESCRIPTION column in the mapping tables, but the clinical file is still using the old values in the detailed columns. If this it the case, then the code would render NAs for the row with mismatch detailed column and DESCRIPTION column. Do we want to throw warning for this?

thomasyu888: Yes, I think it makes sense to surface a warning, although... if there is no match, it becomes

```
ethnicity_mapping$CODE[NA]
```

Does that cause an error?

danlu1: It won't cause an error in this code but I'm not sure if this could break downstream processing. @Chelsea-Na Do you know if we allow NAs in naaccr code columns?

rxu17: So back when I was testing staging workflow for main Genie, I noted that the center JHU had nans in `SECONDARY_RACE`, `TERTIARY_RACE` columns in the production project's `clinical.txt` but Not Collected in those same columns in staging. So there were definitely NAs it seems. 

Even though we have validation [here](https://github.com/Sage-Bionetworks/Genie/blob/develop/genie_registry/clinical.py#L1060-L1106) that should not be letting NAs through in those columns

scripts/case_selection/export_bpc_selected_cases.R
rxu17: Yeah I second what Tom said. I also think for the testing part, yes we will need fake test clinical data for the test cases which is fine for ensuring the matches happening correctly. This is what we do for adding new validation in main Genie as well.

Yes for the integration test part, I can definitely refactor what we have here: https://github.com/Sage-Bionetworks/Genie_processing/blob/main/utility_scripts/run_bpc_table_comparisons.py to run for a single file.

danlu1: Test scripts are added.

scripts/case_selection/export_bpc_selected_cases.R
danlu1: Currently, the code will break if NAs or empty strings are found in naaccr code columns.

rxu17: That's going to be a problem because there are existing nans in the `data_clinical.txt`

thomasyu888: @danlu1 / @rxu17 can we bring the discussion in #genie slack on what we want to do with null secondary and tertiary race values.  The sites impacted are  CHOP, JHU, and PROV.

danlu1: Yes, we can easily convert to throw warning instead of error. "" are in the naaccr code columns. 

Screenshot 2024-09-13 at 11 08 53 AM

thomasyu888 (approved):
🔥 Thanks for all the great work here.  I'll let @rxu17 be the final reviewer - I'm excited about this update.

scripts/case_selection/shared_fxns.R
```diff
@@ -289,11 +289,13 @@ remap_patient_characteristics <- function(clinical, existing_patients, ethnicity
...
 }
 
-#' Check for missing values
+#' Check for missing values in naaccr columns
 #'
 #' @param data The data frame to check against
 #' @param columns The target columns
 check_for_missing_values <- function(data, columns) {
+  # filter out CHOP, PROV, JHU centers with known NAs
+  data <- data[!grepl("CHOP|PROV|JHU", data$genie_patient_id), ]
```
thomasyu888: It's ok to leave it in.  In the future, even after we fix it on main GENIE, there may be issues

danlu1: Once it's fixed on main GENIE, we can drop this row.

thomasyu888: Can we add to the AC of that ticket that rixing created? To drop this row

danlu1: Sure. On it.

scripts/case_selection/shared_fxns.R
```diff
@@ -255,4 +255,55 @@ get_main_genie_clinical_id <- function(release){
...
+  return(patient_df)
+}
+
+#' Check for missing values in naaccr columns
+#'
+#' @param data The data frame to check against
+#' @param columns The target columns
+check_for_missing_values <- function(data, columns) {
+  # filter out CHOP, PROV, JHU centers with known NAs
+  data <- data[!grepl("CHOP|PROV|JHU", data$genie_patient_id), ]
```
rxu17: Could you add a conditional here for filtering out `"CHOP|PROV|JHU"` only if the columns we're checking for missingness for is `naaccr_race_code_tertiary` or `naaccr_race_code_secondary`?

scripts/case_selection/tests/test_shared_fxns.R
```diff
@@ -52,3 +52,103 @@ test_that("get_main_genie_clinical_id returns NULL when data_clinical.txt does n
   result <- get_main_genie_clinical_id(release)
   expect_null(result)
 })
+
+test_that("remap_patient_characteristics works as expected", {
```
rxu17, line 56: Could you add some values to this test case with ""/NA in the naaccr columns?

rxu17 (approved):
LGTM! Just a couple of minor comments

rxu17 (approved):

PR #154, opened 2024-10-01
[GEN-1516, GEN-1555] make table update cohort specific and allow saving files to staging

danlu1:
Problem:

Currently, the update table step is NOT cohort-specific which means that all the data in the internal tables get overwritten. Also, outputs are either saved to prod environment or local. We would like to be able to save them in staging project for testing or validation purpose.

Solution:

1. Table queries are updated to include `cohort` filter.
2. Wipe table line has been updated to wipe cohort lines only.
3. Staging projects are added and Staging tables are used. 

Testing:

Since no new function is added, I think it's better to generate a Comparison report between prod table and staging table after the code is run for a cohort. If tables stay the same, then I think we should be good to go.


scripts/table_updates/update_data_table.py
```diff
@@ -70,9 +76,21 @@ def get_main_genie_clinical_sample_file(
     return clinical_df[["SAMPLE_ID", "SEQ_YEAR"]]
 
 
-def _store_data(syn, table_id, label_data, table_type, logger, dry_run):
+def _store_data(syn, table_id, label_data, table_type, cohort, logger, dry_run):
```
thomasyu888: Nit: add type hints

scripts/table_updates/update_data_table.py
```diff
@@ -104,23 +123,36 @@ def _store_data(syn, table_id, label_data, table_type, logger, dry_run):
...
         if table_type == "primary":
-            table = syn.delete(table_query.asRowSet())  # wipe the table
+            table = syn.delete(table_query)  # wipe the cohort data
         table = syn.store(Table(table_schema, temp_data))
     else:
         temp_data.to_csv(table_id + "_temp.csv")
 
 
-def store_data(syn, master_table, label_data, table_type, logger, dry_run):
+def store_data(syn, master_table, label_data, table_type, cohort, logger, dry_run):
```
thomasyu888: Nit: add type hints

thomasyu888 (approved):
🔥 I took a glance and the logic looks sound to me.  Did you have a chance to run a data comparison report yet between what changed before this vs after? 

I don't see anything in here that would change the value in any way, but it is definitely going to be different that update tables code will no longer update all the cohorts.

Will let @rxu17 give a final review here. Great work!

scripts/table_updates/update_data_table.py
```diff
@@ -104,23 +123,36 @@ def _store_data(syn, table_id, label_data, table_type, logger, dry_run):
     # remove .0 from all columns
     temp_data = temp_data.applymap(lambda x: float_to_int(x))
     # update table
-    table_query = syn.tableQuery("SELECT * from %s" % table_id)
+    table_query = syn.tableQuery(
+        f"SELECT * FROM {table_schema.id} where cohort = '{cohort}'"
+    )
     if table_type == "irr":
```
thomasyu888, line 137: I know we aren't using the irr tables, but by subsetting the tables by `cohort`, if we _do_ run irr, does that mean only a specific cohort is appended?

danlu1: Yes, only the new data for the specific cohort is appended since both the `temp_data` and `existing_records` are cohort specific.

danlu1:
> 🔥 I took a glance and the logic looks sound to me. Did you have a chance to run a data comparison report yet between what changed before this vs after?
> 
> I don't see anything in here that would change the value in any way, but it is definitely going to be different that update tables code will no longer update all the cohorts.
> 
> Will let @rxu17 give a final review here. Great work!

I'm in the process of discussion with Rixing on how to generalize the comparison code because she will also need to tweak the comparison code for her current ticket. Yes, the code changes are mainly regarding controlling the input data and output tables to be cohort specific and adding the functionalities to write to Staging projects.

scripts/table_updates/update_data_table.py
```diff
@@ -399,13 +449,27 @@ def main():
     parser.add_argument(
         "-p", "--project_config", default="config.json", help="Project config file"
     )
+    parser.add_argument(
```
rxu17, line 458: Do we need to update `modules/update_data_table.nf`with these new args? If so, we probably also have to do a nextflow run of this step on staging project to test that that's working.

scripts/table_updates/update_data_table.py
```diff
@@ -218,7 +249,16 @@ def _redact_table(df, interval_cols_info):
     return df, record_to_redact
 
 
-def update_redact_table(syn, redacted_table_info, full_data_table_info, logger):
+def update_redact_table(syn: synapseclient.Synapse, redacted_table_info: pandas.DataFrame, full_data_table_info: pandas.DataFrame, cohort: str, logger: logging.Logger):
+    """Update redacted table
```
rxu17, line 261: Nit: could add more description

scripts/table_updates/update_data_table.py
```diff
@@ -70,9 +76,21 @@ def get_main_genie_clinical_sample_file(
     return clinical_df[["SAMPLE_ID", "SEQ_YEAR"]]
 
 
-def _store_data(syn, table_id, label_data, table_type, logger, dry_run):
+def _store_data(syn: synapseclient.Synapse, table_id: str, label_data: pandas.DataFrame, table_type: str, cohort: str, logger: logging.Logger, dry_run: bool):
+    """Helper function to store data to each table in the master table.
```
rxu17, line 78: Nit: could add more description

rxu17 (approved):
Logic also looks good to me! Had a few comments. Also I think we should also update  [the README](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/develop/scripts/table_updates/README.md) since the parameters changed

scripts/table_updates/update_data_table.py
danlu1: Oh you're right. I forgot to push the change.

modules/update_data_table.nf
```diff
@@ -19,13 +21,17 @@ process update_data_table {
    if (production) {
       """
       cd /root/scripts/
-      python update_data_table.py -p /root/scripts/config.json -m "$comment" primary
+      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -pd
       """
-   }
-   else {
+   } else if (dry_run){
```
rxu17: I was wondering about this for my own ticket - could be a good time for discussion. 

Should we just set a standard for nextflow runs that we don't want to do `dry-run`s for when production mode is on? So `dry-run`s for all future nextflow BPC steps that we modify and add staging for, we only want to have it run for non-production mode. 

But if you really wanted to run `dry-run` for production mode, you technically could locally with the script. 

Thoughts on the above?

danlu1: Ya, `dry-run` might be useful when staging projects are not used. I would say having it in nextflow runs is not absolutely necessary for now. I vote for removing it.

rxu17: Since @Chelsea-Na is usually the one running nextflow runs, I'm wondering if you use `dry-run` for any BPC pipeline runs? If so, since we're adding staging project in so you can toggle between that and production on nextflow, do you still see a use case for using `dry-run` (nothing gets uploaded)?

thomasyu888: Just my 2 cents, sometimes we may want to be able to look at files locally, but we _could_ have that AND upload to staging.  May want to ping Chelsea in a DM, as not sure if her github notifications are working as expected

Chelsea-Na: @rxu17 I agree with the above - if we add staging, I don’t see a need for the `dry-run` option anymore. The main reason I was downloading the output locally was to verify it. Uploading it to a staging project would serve the same purpose.

danlu1: Thanks guys. I will remove the dry-run section.

rxu17: To clarify, we are just removing the dry-run for this nextflow module (not the script itself). We're still keeping dry-run as nextflow parameter because other BPC steps still use it.

scripts/table_updates/utilities.py
```diff
@@ -35,72 +37,123 @@ def float_to_int(val):
...
+
 def check_empty_row(row, cols_to_skip):
     """
     Check if the row of data is empty with given columns to skip
     """
     return row.drop(cols_to_skip).isnull().all()
 
-def download_synapse_table(syn, table_id, condition):
+
+def download_synapse_table(syn, table_id: str, select: str = "*", condition: str = ""):
```
thomasyu888: Nit: add return type hint

thomasyu888 (commented):
This looks great, lets make sure we have Chelsea look at the tests and the data before we push this through!

But the `asDataFrame` solution looks fantastic and allows us to shift to newer versions of python client. We may have to update the Dockerfile @danlu1

modules/update_data_table.nf
```diff
@@ -9,8 +9,10 @@ process update_data_table {
 
    input:
    val previous
-   val comment
+   val cohort
    val production
+   val comment
+   val dry_run
```
thomasyu888: Is the dry_run value used here at all?

thomasyu888 (approved):
🔥 LGTM!  one last comment.  We will most likely have to test this nextflow process somehow after the docker image is built.  We can probably just test the tool individually.

modules/update_data_table.nf
danlu1: Good catch. No. I will remove it.

thomasyu888:
@danlu1 Can this be merged based on the comment from Chelsea in Jira?

danlu1:
@thomasyu888 I didn't check nextflow yet. I just found that I might also need to update [main.nf](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/develop/main.nf). I will test nextflow process after updating main.nf. Once the nextflow check is done, I agree that we can merge the changes.

danlu1:
Rixing's comments on tweaking the test cases for download_synapse_table:

nit: I think it would be helpful to use pytest.mark.parameterize and use their matrix/grid approach to parameters for the test cases
nit: I noticed objects like schema is the same for all of the test cases, could that be a pytest.fixture?
nit: test case naming should be a bit more clear to reflect the differences between test cases: e.g: test_download_synapse_table_no_condition_default_selected_columns for example to represent utilities.download_synapse_table(syn, "syn123456")
Missing test case where select is multiple columns and using default condition (no condition)
nit: Test cases where you're using condition = "" could be revised to just omit that portion because the default for the function is condition = ""

nit: Would rename this test_download_synapse_table_with_selected_column  to be test_download_synapse_table_with_selected_single_column or something like that.

--> Done

PR #155, opened 2024-10-16
[GEN-1485] Modularize update potential PHI fields

rxu17:
**Purpose:** This PR is a draft. Most changes should be present. This PR adds the modularization of the update potential PHI fields code as a nextflow step. This PR also adds the ability to run on staging vs production project. I also added a dry-run ability to the code itself.

**Changes:**
- Instead of having the ability to call the code based on synapse ids, I wanted to be consistent and have the synapse ids be hard coded within the actual code. This way we don't also have hardcoded synapse ids in the nextflow module steps and instead in the code being called itself. Changes [here](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1485-update-potential-phi/scripts/references/update_potential_phi_fields_table.R#L14-L23) and [here](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1485-update-potential-phi/scripts/references/update_potential_phi_fields_table.R#L256-L263).

- Adding [dry-run ability](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1485-update-potential-phi/scripts/references/update_potential_phi_fields_table.R#L238) to the code (not nextflow module). I thought it would be necessarily at first when first developing and testing this code.

- [Added version comment as a parameter](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1485-update-potential-phi/scripts/references/update_potential_phi_fields_table.R#L35-L36) and [provenance](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/gen-1485-update-potential-phi/scripts/references/update_potential_phi_fields_table.R#L240-L245)

-  Minor change but I moved all outside code to a `main` function so this script can be sourced by for example the tests without being called. Also added 

- Added [github actions workflow](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/155/files#diff-2817fc9ec6443e19233164eefa86576a0972ce81bcf0ec4a04f96d6119db038e) to create docker images for the specific branch when there are changes in `scripts/references` for us to use in testing and to ensure that docker builds are successful. I think we can figure out caching in the future (this image only takes ~9 min to build). This can be added to in the future.

**Testing:** Still in progress

- [x] - Running unit tests (still in progress, running into issue with installing `testthat`)
- [x] - Running on staging (still in progress, I'm just copying over my repo changes to the docker image because I'm having issues building it from the Dockerfile)
- [x] - Running [comparison script here](https://github.com/Sage-Bionetworks/Genie_processing/pull/171)
   - No differences found between the updated staging table (v6) and the production table when using the same version of the SOR file. 
   - When comparing between staging table (v2) which is the 2nd to latest version of the production table vs production table, the expected difference was found. More details in [JIRA ticket](https://sagebionetworks.jira.com/browse/GEN-1485)
- [x] - Running nextflow step on staging (this can't happen until we merge the PR otherwise we could create dockerhub tag for this branch)


thomasyu888 (commented):
@danlu1 want to give this a review before I give one?

scripts/references/Dockerfile
rxu17, line 1: Note: This is a fix: https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/2856943651/GENIE+Developer+issues#Issue%3A-No-%2F-outdated-Pub-keys-found-when-building-Docker to an existing issue with the old BPC docker images

.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,67 @@
...
+      id: check_changes
+      run: |
+        # Check for a merge base, fallback to root commit if none exists
+        if git merge-base --is-ancestor origin/develop HEAD; then
+          DIFF_BASE="origin/develop"
+        else
+          DIFF_BASE=$(git rev-list --max-parents=0 HEAD)  # Use the initial commit as fallback
+        fi
+
+        # Compare changes between DIFF_BASE and HEAD
```
danlu1, line 43: Just out of curiosity. How do you envision using the comparison results?

main.nf
```diff
@@ -87,17 +91,25 @@ include { run_clinical_release } from './modules/run_clinical_release'
...
-   merge_and_uncode_rca_uploads(run_quac_upload_report_warning.out, ch_cohort, params.production)
-   // remove_patients_from_merged(merge_and_uncode_rca_uploads.out, ch_cohort, params.production)
-   update_data_table(merge_and_uncode_rca_uploads.out, ch_comment, params.production)
-   update_date_tracking_table(update_data_table.out, ch_cohort, ch_comment, params.production)
-   run_quac_table_report(update_date_tracking_table.out, ch_cohort, params.production)
-   run_quac_comparison_report(run_quac_table_report.out, ch_cohort, params.production)
-   create_masking_report(run_quac_comparison_report.out, ch_cohort, params.production)
-   update_case_count_table(create_masking_report.out, ch_comment, params.production)
+   
+   if (params.step == "update_potential_phi_fields_table") {
```
danlu1, line 95: I feel like testing only `update_potential_phi_fields_table` step can be done in its specific module. Or we can surface this step as a separate pipeline as https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/genie-bpc-project/launchpad/272225815007836

scripts/references/Dockerfile
```diff
@@ -1,21 +1,34 @@
...
+    rm -rf /var/lib/apt/lists/*
 
-ENV PYTHON /usr/local/lib/R/site-library/PythonEmbedInR/bin/python3.6
+# Install R packages including remotes and renv
+RUN R -e "install.packages('remotes', repos = 'https://cloud.r-project.org')" && \
+    R -e "remotes::install_github('rstudio/renv', ref = '${RENV_VERSION}')" || true
 
-RUN R -e "install.packages('remotes', repos = c(CRAN = 'https://cloud.r-project.org'))"
-RUN R -e "remotes::install_github('rstudio/renv@${RENV_VERSION}')"
+# Install synapser with specific version
```
danlu1, line 21: I was just curious if there was a specific reason for sticking with that 0.11.7 and PythonEmbededInR instead of updating to the latest synapser? I'm wondering if there's something I should be aware of, like compatibility issues with the new version.

danlu1 (commented):
Nice job. Leave a few comments.

main.nf
rxu17: Discussed this but you can set `params.step` to a specific step and run the workflow for just that step.

.github/workflows/build-docker-images.yml
rxu17: I think the comparison results will come into play when we add in other docker images because we don't want to re-build docker images for all when we just modify the code for one of them. Right now I only have one for `scripts/references` but it did work when I pushed edits to this GH actions workflow file and it skipped the build for the docker image.

danlu1: why don't remove path '.github/workflows/build-docker-images.yml'?

danlu1: I would prefer to have this workflow for other repos as well. So we can test changes without needing to manually update corresponding images.

rxu17: Mainly keeping it because maybe we'll want the workflow to run even when we don't have changes to docker

scripts/references/Dockerfile
rxu17: No reason other than we already know the code works with this docker environment. Also because updating R env can be quite painful.

scripts/references/update_potential_phi_fields_table.R
```diff
@@ -215,57 +224,95 @@ get_red_table_update <-function(var_add, inf_sor) {
...
+
+  toc = as.double(Sys.time())
+  print(glue("Runtime: {round(toc - tic)} s"))
 }
 
-toc = as.double(Sys.time())
-print(glue("Runtime: {round(toc - tic)} s"))
+
+# only run main when not sourced
+if (sys.nframe() == 0) {
```
thomasyu888, line 316: Nice! I didn't know this was possible!

thomasyu888 (approved):
🔥 LGTM! Pre-approving - will leave it to Dan to be final reviewer.

.github/workflows/build-docker-images.yml
danlu1: okay. i thought we only need this to run when changes are made to the reference repo.

danlu1 (approved):

PR #156, opened 2024-10-18
[GEN-1468] overwrite tier1 variable

danlu1:
Problem:

The tier1a variables (race, sex, ethnicity, sample_type, seq_date) in BPC tables need to be replaced with values extracted from Main GENIE.

Solution:

1. Updated get_main_genie_clinical_sample_file to allow it more generic to pull both patient and sample release files. 
2. Added update_tier1a and overwrite_tier1a functions to update BPC tables allowing to update for all cohorts or a specific cohort. 

Testing:

Unit tests have been added.


scripts/table_updates/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.9
```
thomasyu888: Nit: May want to consider just bumping it to Python 3.10 or 3.11 since Python 3.9 is EOL next year.

scripts/table_updates/config.json
```diff
@@ -19,7 +19,7 @@
         "CRC2": "syn52943210",
         "RENAL": "syn59474249"
     },
-    "main_genie_release_version": "16.6-consortium",
+    "main_genie_release_version": "17.4-consortium",
```
thomasyu888: @Chelsea-Na do we want to keep this at 17.2?

scripts/table_updates/update_data_table.py
```diff
@@ -234,16 +307,11 @@ def update_redact_table(syn, redacted_table_info, full_data_table_info, logger):
...
-    patient_info = syn.tableQuery(
-        "SELECT record_id, birth_year, hybrid_death_ind FROM %s" % patient_table_id
-    ).asDataFrame()
-    sample_info = syn.tableQuery(
-        "SELECT record_id, cpt_genie_sample_id, age_at_seq_report FROM %s"
-        % sample_table_id
-    ).asDataFrame()
+    # download tables
+    condition = f"cohort = '{cohort}'"
+    curation_info = utilities.download_synapse_table(syn, curation_table_id, "record_id, curation_dt", condition)
```
thomasyu888: Nit: Might be helpful to have 

```download_synapse_table(syn=syn, arg2=..., arg3...)```

So we know what the parameters are.

File: scripts/table_updates/utilities.py [hunk 4]
```diff
@@ -126,12 +186,95 @@ def revert_table_version(syn, table_id):
...
+        cpt_table_id (str): The synapse ID for BPC table to be modified
+        cpt_seq_dat (pandas.DataFrame): The updated BPC table as dataframe
+        bpc_column_list (List[str]): The column list to be replaced
+        logger (logging.Logger, optional): The custom logger. Defaults to None.
+    """
+    # check the validity of bpc_column_list
+    logger.info(f"Overwrite {bpc_column_list} in {form}")
+    # load bpc table
+    cpt_table_schema = syn.get(cpt_table_id)
+    cpt_dat_query = syn.tableQuery(f"SELECT * FROM {cpt_table_id}")
```
thomasyu888: Does this need to take into account the cohort?

scripts/table_updates/update_data_table.py
danlu1: Sure. I will add.

scripts/table_updates/utilities.py
danlu1: Not necessary. I've tested it. Whether to include condition in the table query or not gives the identical etag. Here is a screenshot of my testing.

Screenshot 2024-10-19 at 12 56 55 AM

scripts/table_updates/Dockerfile
danlu1: I put 3.9 here to follow what is specified in synapseclient dockerfile. I'm happy to bump it up to a newer version.

thomasyu888: Hmm, I'll let you decide - I think this works for now.

scripts/table_updates/config.json
Chelsea-Na: Good question! If we are ready to test the output, we should test it out on 17.4-consortium.  We will eventually need to also test on 17.6-consortium once its out. 

Do we know what happens if a main GENIE value is missing? Or if the sample/patient is missing? Or if it adds to a log if there is a mismatch between the upload and the replaced value?

scripts/table_updates/utilities.py
```diff
@@ -35,72 +39,123 @@ def float_to_int(val):
...
+        A Pandas dataframe of the Synapse table
     """
     if condition:
-        condition = " WHERE "+condition
-    synapse_table = syn.tableQuery("SELECT * from %s%s" % (table_id,condition))
-    synapse_table = synapse_table.asDataFrame()
-    return(synapse_table)
+        condition = " WHERE " + condition
+    synapse_table = syn.tableQuery(f"SELECT {select} from {table_id}{condition}")
+    na_values = [
```
rxu17: Nit: Can this be a global variable or function so you can pull from it in the various places you need it for?

scripts/table_updates/utilities.py
```diff
@@ -35,72 +39,123 @@ def float_to_int(val):
...
-def download_synapse_table(syn, table_id, condition):
+
+def download_synapse_table(syn, table_id: str, select: str = "*", condition: str = "") -> pandas.DataFrame:
     """Download Synapse Table with the given table ID and condition
-    
+
     Args:
         syn: Synapse credential
         table_id: Synapse ID of a table
+        select: Columns to be selected
```
rxu17: Nit: Add explanation for the default value which is *: Default: all columns

scripts/table_updates/config.json
danlu1: @Chelsea-Na This is a hard replacement so I didn't check GENIE value missingness. I checked with Rixing that we think all BPC samples/patients should be in Main GENIE referring to [BPC project description](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/2163344270/Onboarding#GENIE-BPC:~:text=Launched%20in%202019%2C%20this%20is%20an%20augmentation%20of%20the%20main%20GENIE%20project%20that%20adds%20in%20manually%20curated%20clinical%20data%20elements%20of%20a%20subset%20of%20previously%20collected%20main%20GENIE%20cases%20using%20the%20PRISSMM%20Data%20Model.). I'm happy to add a check for sample/patient missingness.  Moreover, it doesn't log for the discrepancies between uploaded and replaced values. I'm happy to add a function for it if it's helpful.

scripts/table_updates/update_data_table.py
danlu1: Note, I will address this in the PR for GEN-1516 since it will be merged first.

scripts/table_updates/tests/test_utilities.py
```diff
@@ -0,0 +1,249 @@
...
+import synapseclient
+from synapseclient import Schema, Table, client
+from table_updates import utilities
+
+
+@pytest.fixture(scope="session")
+def syn():
+    return create_autospec(synapseclient.Synapse)
+
+def test_download_synapse_table_with_selected_columns(syn):
```
rxu17: nit: Would rename this one to be `test_download_synapse_table_with_selected_single_column` or something like that.

scripts/table_updates/utilities.py
```diff
@@ -126,12 +186,95 @@ def revert_table_version(syn, table_id):
...
+    table_columns = list(set(table_columns) & set(old_data.columns))
     temp_data = old_data[table_columns]
     table_query = syn.tableQuery("SELECT * from %s" % table_id)
     syn.delete(table_query.asRowSet())
-    syn.store(Table(table_schema, temp_data))
\ No newline at end of file
+    syn.store(Table(table_schema, temp_data))
+
+
+def update_tier1a(syn: synapseclient.Synapse, form: str, master_table: pandas.DataFrame, main_genie_table: pandas.DataFrame, column_mapping_table: pandas.DataFrame, bpc_column_list: List[str],logger: logging.Logger = None, cohort: str = "") -> Tuple[str, pandas.DataFrame]:
```
rxu17: Nit: do you forsee this function and `overwrite_tier1a` being used in other scripts under scripts/table_updates? It seems like a specific function specific to `update_data_table.py` rather than a general utilities function?

scripts/table_updates/tests/test_utilities.py
rxu17, line 1: Some thoughts on the test cases for `download_synapse_table`:
- **nit:** I think it would be helpful to use `pytest.mark.parameterize` and use their matrix/grid approach to parameters for the test cases
- **nit:** I noticed objects like `schema` is the same for all of the test cases, could that be a pytest.fixture? 
- **nit:** test case naming should be a bit more clear to reflect the differences between test cases: e.g: `test_download_synapse_table_no_condition_default_selected_columns` for example to represent `utilities.download_synapse_table(syn, "syn123456")`
- Missing test case where `select` is multiple columns and using default condition (no condition)
- **nit:** Test cases where you're using `condition = ""` could be revised to just omit that portion because the default for the function is `condition = ""`

scripts/table_updates/tests/test_utilities.py
```diff
@@ -0,0 +1,249 @@
...
+        main_genie_table = MagicMock()
+
+        utilities.update_tier1a(syn, form, master_table, main_genie_table, column_mapping_table, bpc_column_list, mock_logger)
+
+        # validate
+        assert str(excinfo.value) == expected_error
+        mock_logger.assert_not_called()
+
+
+@pytest.mark.parametrize(
```
rxu17: nit: there's a lot of parameters here so it makes it really hard to read when using `pytest.mark.parametrize` (usually once I have more than 3 parameters, I'd use a different method). I'd recommend something like [this](https://github.com/Sage-Bionetworks/Genie/blob/6ec4fe1f9e667e69b1f2e21553473a260d2e53bb/tests/test_transform.py#L192-L238)

rxu17 (commented):
Just did a first pass, had some comments

scripts/table_updates/utilities.py
danlu1: My original design is that we use `download_synapse_table` as the default function whenever we want to read Synapse tables. I also update lines calling `asDataFrame` to call the `download_synapse_table`.  I put na_values within the function since I prefer this function can be used directly without loading NA list when it is called.

scripts/table_updates/utilities.py
danlu1: I will address this in the PR for GEN-1516 since it will be merged first. Also, this is a function added in this [PR](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/154).

scripts/table_updates/utilities.py
danlu1: I think it's a function specific to update_data_table.py.

scripts/table_updates/utilities.py
danlu1: Or do you mean make na_values a global variable in the utilities.py?

scripts/table_updates/tests/test_utilities.py
danlu1: Will address this in this [PR](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/154).

scripts/table_updates/Dockerfile
danlu1: Tested it that the docker image can be built successfully for python 3.11. I will include this change in GEN-1516.

scripts/table_updates/utilities.py
rxu17: > Or do you mean make na_values a global variable in the utilities.py?

This, since it just gets used in utilities.py

scripts/table_updates/utilities.py
rxu17: Okay at some point it might make sense to move it to `update_data_table.py` but again just a nit

scripts/table_updates/utilities.py
```diff
@@ -126,12 +186,95 @@ def revert_table_version(syn, table_id):
...
+        column_mapping_table (pandas.DataFrame): The column mapping table between BPC and Main GENIE
+        bpc_column_list (List[str]): The column list to be replaced
+        logger (logging.Logger, optional): The custom logger. Optional. Defaults to None.
+        cohort (str, optional): The cohort name. Defaults to "".
+
+    Returns:
+        Tuple[str, pandas.DataFrame]: The synapse ID for BPC table to be modified and the updated table as dataframe
+    """
+    # check the validity of bpc_column_list
+    valid_col = column_mapping_table.loc[column_mapping_table["prissmm_form"] == form,].prissmm_element.tolist()
```
rxu17: Why are we using `prissm_form`? Could there be somewhere in the function docstring describing why we are pulling our column list for both bpc and main genie from here?

scripts/table_updates/utilities.py
```diff
@@ -126,12 +186,95 @@ def revert_table_version(syn, table_id):
...
+    if form == "patient_characteristics":
+        main_genie_table = main_genie_table[main_genie_column_list + ["PATIENT_ID"]]
+        cpt_seq_dat = cpt_dat.merge(
+            main_genie_table,
+            how="left",
+            left_on="genie_patient_id",
+            right_on="PATIENT_ID",
+        )
+    else: 
+        main_genie_table = main_genie_table[main_genie_column_list + ["SAMPLE_ID"]]
```
rxu17: I'm not sure what the original code looked like but was there ever a handling of potential duplicates before (when we first query by cohort) and after merging here?

scripts/table_updates/utilities.py
```diff
@@ -126,12 +186,95 @@ def revert_table_version(syn, table_id):
...
+        cpt_seq_dat = cpt_dat.merge(
+            main_genie_table,
+            how="left",
+            left_on="cpt_genie_sample_id",
+            right_on="SAMPLE_ID",
```
rxu17: I see Chelsea's concern here: what happens if there isn't a 1:1 merge between bpc and main genie (BPC has sample/patients not present in clinical). How did the code previously handle the merge?

danlu1: The original code doesn't handle duplicate. See here: https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/1bc58ec5c7415ba5b989dbb5a0de39b4839a1b0b/scripts/table_updates/update_data_table.py#L346C1-L357C6

scripts/table_updates/utilities.py
danlu1: The reason is the prissmm_form matches form column in the [Data Table information table](https://www.synapse.org/Synapse:syn63616766/tables/). I can update the doctring.

scripts/table_updates/utilities.py
danlu1: The original code doesn't handle duplicate.

scripts/table_updates/utilities.py
danlu1: Done.

scripts/table_updates/config.json
danlu1: I forgot to address your question: "Do we know what happens if a main GENIE value is missing? Or if the sample/patient is missing? 
If the main GENIE sample/patient is missing, the values will be filled with `NaN`. 
If the GENIE value is missing, then the final dataframe shows missingness since we use all main GENIE values to hard replace BPC. However, this might not be the case because no missingness has been found in target columns in the [sample clinical table](https://www.synapse.org/Synapse:syn7517674/tables/) and [patient clinical table](https://www.synapse.org/Synapse:syn7517669/tables/).

danlu1: @Chelsea-Na Two functions have been added to check for sample/patient missingness and log the discrepancies between uploaded and replaced values.

scripts/table_updates/tests/test_utilities.py
danlu1: Done.

scripts/table_updates/update_data_table.py
```diff
@@ -423,79 +454,258 @@ def update_redact_table(
...
+        cohort (str, optional): The cohort name. Defaults to "".
+
+    Returns:
+        Tuple[str, pandas.DataFrame]: The synapse ID for BPC table to be modified and the updated table as dataframe
+    """
+    # check the validity of bpc_column_list
+    valid_col = column_mapping_table.loc[
+        column_mapping_table["prissmm_form"] == form,
+    ].prissmm_element.tolist()
+    assert all(
```
thomasyu888: Nit: assertions are used for tests, I would raise a ValueError here.

scripts/table_updates/update_data_table.py
```diff
@@ -423,79 +454,258 @@ def update_redact_table(
...
+        )
+        cpt_seq_dat = cpt_dat.merge(
+            main_genie_table,
+            how="left",
+            left_on="cpt_genie_sample_id",
+            right_on="SAMPLE_ID",
+        )
+        utilities.update_tier1a_data_replacement_mapping_table(
+            syn, cpt_seq_dat, form, config
+        )
```
thomasyu888: Nit: there is duplication of code here, but it's minimal.

thomasyu888 (approved):
🔥 Great work here!  I'll defer to @rxu17 for a final review!

scripts/table_updates/config.json
thomasyu888: Let's push the version to 17.6-consortium

scripts/table_updates/config.json
```diff
@@ -18,10 +18,14 @@
         "BLADDER": "syn26721151",
         "NSCLC2": "syn51318736",
         "CRC2": "syn52943210",
-        "RENAL": "syn59474249"
+        "RENAL": "syn59474249",
         "OVARIAN": "syn64042773"
     },
-    "main_genie_release_version": "16.6-consortium",
+    "tier1a_replacement_mapping":{
+        "patient_characteristics_tier1a_replacement_mapping_table": "syn64331052",
```
thomasyu888, line 25: May I ask what these tables are for?

danlu1: The `patient_characteristics_tier1a_replacement_mapping_table` and `cancer_panel_test_tier1a_replacement_mapping_table` track how original bpc codes are mapped to NAACCR code. This is added to address Chelsea's request: "adds to a log if there is a mismatch between the upload and the replaced value".

scripts/table_updates/update_data_table.py
```diff
@@ -37,42 +50,56 @@
...
+        ]["fileSynId"].values[0]
+    else:
+        clinical_link_synid = release_files[
+            (release_files["release"] == release)
+            & (release_files["name"] == "data_clinical_sample.txt")
+        ]["fileSynId"].values[0]
     clinical_ent = syn.get(clinical_link_synid, followLink=True)
     clinical_df = pandas.read_csv(clinical_ent.path, sep="\t", skiprows=4)
+    # get column list for the form
+    column_list = column_mapping_table.loc[
```
rxu17, line 82: The column mapping table is a fileview right?

scripts/table_updates/update_data_table.py
```diff
@@ -423,79 +454,258 @@ def update_redact_table(
...
+        right_on (str): The join key of the right table
+        logger (logging.Logger, optional): The custom logger. Optional. Defaults to None.
+    Returns:
+        logging.Logger: warning message if join key is not available in the right table
+    """
+    missing_join_keys = left_table.loc[
+        ~left_table[left_on].isin(right_table[right_on])
+    ][left_on]
+    if len(missing_join_keys) > 0:
+        logger.warning(
```
rxu17, line 472: Hmm have you run into this before where main genie patient ids may not match BPC patient ids?

scripts/table_updates/update_data_table.py
```diff
@@ -423,79 +454,258 @@ def update_redact_table(
...
+    # unlist form column in master table
+    master_table["form"] = master_table["form"].apply(lambda x: ", ".join(x))
+    # modify for patient table
+    cpt_table_id, cpt_seq_dat = update_tier1a(
+        syn,
+        "patient_characteristics",
+        master_table,
+        genie_patient_dat,
+        column_mapping_table,
+        bpc_column_list=[
```
rxu17: nit: could make this a list that gets passed in incase this list changes

scripts/table_updates/update_data_table.py
```diff
@@ -423,79 +454,258 @@ def update_redact_table(
...
+            main_genie_table,
+            how="left",
+            left_on="cpt_genie_sample_id",
+            right_on="SAMPLE_ID",
+        )
+        utilities.update_tier1a_data_replacement_mapping_table(
+            syn, cpt_seq_dat, form, config
+        )
+    # reformat the columns
+    cpt_seq_dat.index = cpt_seq_dat["index"]
```
rxu17, line 568: Since `cpt_seq_dat` goes through a bunch of formatting here and I don't see tests for `update_tier1a` function, might be worth putting these in a separate function and testing that. I don't know how much this section of code changed from previous

rxu17 (approved):
LGTM overall! Some comments/questions

scripts/table_updates/update_data_table.py
danlu1: Okay. I will also tweak old code to remove assertions.

scripts/table_updates/update_data_table.py
danlu1: Nope. I added this to address Chelsea's and your question: "Do we know what happens if a main GENIE value is missing? Or if the sample/patient is missing?" + "what happens if there isn't a 1:1 merge between bpc and main genie (BPC has sample/patients not present in clinical)".

scripts/table_updates/update_data_table.py
danlu1: No, it is a table: https://www.synapse.org/Synapse:syn20945902/tables/

scripts/table_updates/update_data_table.py
rxu17: I wonder if we should just enforce an error here. We might miss the logging. But if it was happening, it doesn't sound like a big issue.

danlu1: I think warning might makes the most sense in this context since we just wanted to be informed that some of the tier1a fields will be filled with NaN.

scripts/table_updates/update_data_table.py
danlu1: Oh this is because I forgot to push the test scripts for update_tier1a and will push it with other changes.

scripts/table_updates/update_data_table.py
danlu1: Done.

PR #157, opened 2024-10-18
Add ghcr workflow

rxu17:
**Purpose:** Adds github actions workflow (using matrix method) for GHCR docker builds

Can ignore failed docker build of `references` modules because it is fixed in this PR #155 

Depends on #155

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -1,0 +1,67 @@
+name: Build and Push Docker Images
+
+on:
+  push:
+    branches: [develop, 'GEN*', 'gen*']
+    paths:
+      - 'scripts/references/**'
+      - '.github/workflows/build-docker-images.yml'
+  workflow_dispatch:
+
+jobs:
+  build_references_docker:
+    runs-on: ubuntu-latest
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: sage-bionetworks/genie-bpc-pipeline
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+        
+    - name: Setup Docker buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Fetch the default branch (develop) for comparison
+      run: git fetch origin develop:refs/remotes/origin/develop --depth=1
+
+    - name: Check for Changes in scripts/references
+      id: check_changes
+      run: |
+        # Check for a merge base, fallback to root commit if none exists
+        if git merge-base --is-ancestor origin/develop HEAD; then
+          DIFF_BASE="origin/develop"
+        else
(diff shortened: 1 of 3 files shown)


.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,67 @@
...
+      - 'scripts/references/**'
+      - '.github/workflows/build-docker-images.yml'
+  workflow_dispatch:
+
+jobs:
+  build_references_docker:
+    runs-on: ubuntu-latest
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: sage-bionetworks/genie-bpc-pipeline
```
thomasyu888, line 19: Just a quick note: did we want the different folders to be tags OR following the same structure as we have now in dockerhub with

...-pipeline-case-selection:develop

danlu1: i guess this is achieved in the Build and Push Docker Image step. ```tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:references-${{ github.ref_name }}```

rxu17: Right now the structure is `<registry>/<repo>:<folder_name>-<branch>` (example: `genie-bpc-pipeline:references-gen-1485-update-potential-phi`) mainly because in GHCR, I don't believe we can do `<registry>/<repo>-<folder_name>:-<branch>`. I tried to concatenate to the `repo` part of the name but GHCR doesn't allow you to override the repo name from my understanding. So I have it attached to the tag. 

But yes have a tag for each branch so we can do dev work and not break the current working image.

thomasyu888 (approved):
🔥 LGTM!  Going to pre-approve

.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,70 @@
+name: Build and Push Docker Images
+
+on:
+  push:
+    branches: [develop, 'GEN*', 'gen*']
+    paths: # add the paths to each new module
+      - 'scripts/references/**'
```
danlu1: Change this to scripts/**

danlu1 (approved):
Excellent job. I will test the new docker image for my changes to the table_updates module.

.github/workflows/build-docker-images.yml
```diff
@@ -30,18 +33,24 @@ jobs:
...
       run: git fetch origin develop:refs/remotes/origin/develop --depth=1
 
-    - name: Check for Changes in scripts/references
+    - name: Check for Changes in scripts/${{ matrix.module }}
       id: check_changes
       run: |
-        # Check for a merge base, fallback to root commit if none exists
-        if git merge-base --is-ancestor origin/develop HEAD; then
-          DIFF_BASE="origin/develop"
+        # Determine the correct DIFF_BASE
```
rxu17, line 39: Added logic to determine changes against develop (when doing a merge) correctly. I tested it with a local merge on develop and used these commands (it seemed to work).

PR #159, opened 2024-10-27
[GEN-1577] Update README.md to help with installation of running the scripts

thomasyu888:
Update README.md to help with installation of running the scripts

### DIFF
--- a/scripts/table_updates/README.md
+++ b/scripts/table_updates/README.md
@@ -22,9 +22,6 @@ Please make sure you have the [.synapseConfig file](https://help.synapse.org/doc
 Make sure you have Python 3.8 and conda installed.
 
 ### Install the required packages
-```
-pip install 'synapseclient[pandas] == 2.7.2'
-```
 
 > [!NOTE]  
 > Due to this tool using an older version of the python client, until there is bandwidth to update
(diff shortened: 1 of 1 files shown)


Chelsea-Na (approved):

scripts/table_updates/README.md
```diff
@@ -6,11 +6,14 @@ BPC Table Update
 Installation and Setup
 ----------------------
 ### Python version
-Make sure you have Python 3.8 installed
+Make sure you have Python 3.8 installed.
```
danlu1: I think we should update this to be 3.9 and above to be compatible with Python client.

scripts/table_updates/README.md
```diff
@@ -20,6 +23,15 @@ Use a t3.2xlarge ec2 instance for large memory requirement.
 
 Putting it all together.
 
+> [!NOTE]  
+> Due to this tool using an older version of the python client, until there is bandwidth to update
```
danlu1, line 27: The Python client dependency has been updated to be 4.6.0.

danlu1 (changes requested):

scripts/table_updates/README.md
thomasyu888: Hmm, we may need to split off this script in the interim, because there is a specific script in here that may not work with python 4.6

scripts/table_updates/README.md
thomasyu888: Ill let you merge in your PR first, then Ill make some changes to this PR.  This is specific to a script that Chelsea is trying to run

PR #161, opened 2024-10-28
[GEN-1574] Add merge and decode to staging

rxu17:
**Purpose:** We currently don't have a way to safely test and upload results to non-production environment for `merge_and_decode` step in genie bpc pipeline. This is to add that.

**Changes:**
Aside from the nextflow workflow changes, here are the relevant changes in the `merge_and_uncode_rca_uploads.R` script:

- New helper function [get_output_folder_id](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/161/files#diff-f31d1ddec922f34eb2145359628decffa51aa77c780c287f3d1862b4a49a3b5fR638-R644) to be used to retrieve the output synapse id
- New [`production` flag](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/161/files#diff-f31d1ddec922f34eb2145359628decffa51aa77c780c287f3d1862b4a49a3b5fR705-R706) and logic surrounding it to retrieve the correct entities for staging vs production
- New [comment flag](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/pull/161/files#diff-f31d1ddec922f34eb2145359628decffa51aa77c780c287f3d1862b4a49a3b5fR709-R710) NOTE: this is just a placeholder, we will need to update R and synapser drastically to be able to use this, and that's beyond the scope of this ticket.

**Testing:**
- Added unit testing for helper function `get_output_folder_id`
- Tested on staging project (locally and on tower). Comparison results for a couple of cohorts saved to [JIRA ticket](https://sagebionetworks.jira.com/browse/GEN-1574)
- Not a perfect integration test but ran on `develop` branch docker (except changed the synapse id of the output to staging) and compared it to results on the feature branch docker and no differences found!

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        module: ["references", "table_updates"] # Define the modules you want to loop through for builds
+        module: ["references", "table_updates", "uploads"] # Define the modules you want to loop through for builds
     env:
       REGISTRY: ghcr.io
       IMAGE_NAME: sage-bionetworks/genie-bpc-pipeline
--- a/scripts/uploads/merge_and_uncode_rca_uploads.R
+++ b/scripts/uploads/merge_and_uncode_rca_uploads.R
@@ -487,12 +487,22 @@ get_irr <- function(data) {
   return(irr)
 }
 
-save_to_synapse <- function(path, parent_id, file_name = NA, prov_name = NA, prov_desc = NA, prov_used = NA, prov_exec = NA) {
+#' Stores the file, version comment and provenance to Synapse
+#' @param path (string) name of cohort
+#' @param parent_id (string) whether we are running in production env or staging env
+#' @param comment (string) some sort of comment about the new version of the file 
+#' @param file_name (string) file name
+#' @param prov_name (string) name of provenance
+#' @param prov_desc (string) provenance description
+#' @param prov_used (string) url/link to provenance used
+#' @param prov_exec (string) url/link to what was used to execute provenance
+save_to_synapse <- function(path, parent_id, comment, file_name = NA, prov_name = NA, prov_desc = NA, prov_used = NA, prov_exec = NA) {
   
   if (is.na(file_name)) {
     file_name = path
   } 
   file <- File(path = path, parentId = parent_id, name = file_name)
+  file$properties$versionComment <- comment
   
   if (!is.na(prov_name) || !is.na(prov_desc) || !is.na(prov_used) || !is.na(prov_exec)) {
     act <- Activity(name = prov_name,
@@ -635,7 +645,9 @@ get_output_folder_id <- function(config, environment){
 #' Remove leading and trailing whitespace from a string.
 #' @param cohort (string) name of cohort
 #' @param environment (string) whether we are running in production env or staging env
-save_output_synapse <- function(cohort, environment) {
+#' @param comment (string) some sort of comment about the new version of the file 
+#'  related to cohort run
+save_output_synapse <- function(cohort, environment, comment) {
   
   parent_id <- get_output_folder_id(config, environment)
   file_output_pri <- get_pri_file_name(cohort)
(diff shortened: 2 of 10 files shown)


scripts/uploads/merge_and_uncode_rca_uploads.R
```diff
@@ -535,11 +508,13 @@ save_to_synapse <- function(path, parent_id, file_name = NA, prov_name = NA, pro
...
                     used = prov_used,
                     executed = prov_exec)
-    file <- synStore(file, activity = act)
+    file_id <- synStore(file, activity = act)$properties$id
   } else {
-    file <- synStore(file)
+    file_id <- synStore(file)$properties$id
   }
-  
+  file_no_comment <- synGet(file_id, downloadFile=FALSE)
```
danlu1: As my understanding, this would generate a new version of the file with the comment. I suggest add the comment before store the file. It could be something like: 
```
file$properties$versionComment <- comment
file_id <- synStore(file, activity = act)$properties$id
```

scripts/uploads/merge_and_uncode_rca_uploads.R
```diff
@@ -523,7 +487,16 @@ get_irr <- function(data) {
   return(irr)
 }
 
-save_to_synapse <- function(path, parent_id, file_name = NA, prov_name = NA, prov_desc = NA, prov_used = NA, prov_exec = NA) {
+#' Stores the file, version comment and provenance to Synapse
+#' @param path (string) name of cohort
```
danlu1: I guess this should be the intake data file path.

scripts/uploads/merge_and_uncode_rca_uploads.R
```diff
@@ -523,7 +487,16 @@ get_irr <- function(data) {
...
+#' @param parent_id (string) whether we are running in production env or staging env
```
danlu1: I see your point that you want to flag that the parent_id should be the folder in either production or staging proejcts. But I think we should also point out that it should be a folder synapse id.

danlu1 (changes requested):
Great work. Leave a few minor comments.

scripts/uploads/merge_and_uncode_rca_uploads.R
rxu17: Oops thanks for the catch. I think my comments got scrambled up in this function.

scripts/uploads/merge_and_uncode_rca_uploads.R
rxu17: Agreed, I'm going to just go back to the original method (as placeholder)

danlu1 (approved):
LTGM!

.github/workflows/build-docker-images.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        module: ["references", "table_updates"] # Define the modules you want to loop through for builds
+        module: ["references", "table_updates", "uploads"] # Define the modules you want to loop through for builds
```
thomasyu888, line 16: This is wonderful! Yay for scalable solutions!

scripts/uploads/merge_and_uncode_rca_uploads.R
```diff
@@ -698,90 +685,147 @@ save_output_synapse <- function(cohort) {
...
+  cohorts <- names(config$upload)
+  cohorts_str <- paste0(cohorts, collapse = ", ")
+
+  option_list <- list( 
+    make_option(c("-c", "--cohort"), type = "character",
+                help=glue("BPC cohort (choices: {cohorts_str}, {config$misc$all})")),
+    make_option(c("-u", "--save_synapse"), action="store_true", default = FALSE, 
+                help="Save output to Synapse"),
+    make_option(c("-a", "--synapse_auth"), type = "character", default = NA,
+                help="Path to .synapseConfig file or Synapse PAT (default: normal synapse login behavior)"),
```
thomasyu888, line 701: Nit: don't need to do this now, but in the future if you see these, it can all be removed since we added support SYNAPSE_AUTH_TOKEN style login to the synapse client.

thomasyu888 (approved):
🔥 LGTM!

PR #162, opened 2024-11-11
[GEN-1240] Add grs toggle

rxu17:
**Purpose:** This PR is a draft PR. Majority of the code design is present. At times we would like to use the data dictionary (dd) as the primary mapping for a dataset over the combined mapping of dd and global response set (grs) for particular cohorts. This PR adds the ability to specify a `use_grs` parameter for whether we should use grs or not for the cohort. 

This is specific only to the `merge_and_decode` step. This is the 1st PR in a series. See [2nd PR in series](https://github.com/Sage-Bionetworks/GENIE-Sponsored-Projects/pull/141)

**Changes:**

- Added `production` and `use_grs` parameters and logic to support that
- Added helper functions `get_global_response_set` and `get_prov_used` for easier testability and incorporation of the `use_grs` param

**Testing:**
- [x] Unit tests (still in progress)
- [x] Run `merge_and_decode` step with `use_grs` on staging and compare to production run in develop (should be the same result) 
- [x] Run `merge_and_decode` step **without** `use_grs` and compare to production run in develop (should have mapping differences)


main.nf
```diff
@@ -25,6 +25,7 @@ params.production = false
 params.schema_ignore_params = ""
 params.help = false
 params.step = "update_potential_phi_fields_table"
+params.use_grs = false
```
thomasyu888, line 28: Will want to let Chelsea know that she will have to set `use_grs` to true for some of the cohort release we have to do in the coming months.

main.nf
```diff
@@ -25,6 +25,7 @@ params.production = false
 params.schema_ignore_params = ""
```
thomasyu888, line 25: Nit: is there way to limit the boolean values to be boolean.  For example, I think chelsea has said that accidental mis-spellings of production or use_grs will trigger production pipeline.

thomasyu888 (approved):
🔥 LGTM , with minor comments

main.nf
rxu17: yes will add to SOP once merged

main.nf
rxu17: Could add checks, I'll look into it for nextflow

rxu17: Hmm i think this is avoidable now because we have drop down options for all of these parameters in Tower whereas before for the bpc pipeline launch that wasn't available.. You no longer need to type it out. 
image

PR #164, opened 2024-12-09
[GEN-1636] add IDs from retraction form in the retraction steps for creating clinical release files

xindiguo:


scripts/release/update_retraction_table.py
```diff
@@ -1,208 +1,243 @@
...
 
 def update_version(syn, table_id, comment):
+    """Update the version of the table with a new snapshot and comment.
+    
+    Args:
+        syn (Synapse): Synapse client object
+        table_id (str): Synapse Table ID
+        comment (str): Comment to attach to the snapshot
     """
-    Update the table version with given table ID and comment
```
thomasyu888, line 77: I actually think there is a syn.create_snapshot function that we could look into here

thomasyu888 (commented):
I glanced at this and the overall logic looks good, but I'll leave it to @rxu17 for final review.

We should add tests to any new code that is written even if we don't have CI/CD set up.

scripts/release/update_retraction_table.py
rxu17: Yup, [here is an example](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/d05e92749706e464af9ea0f3797ffc1f318c272b/scripts/references/update_potential_phi_fields_table.R#L146-L153)

scripts/release/create_release_files.R
```diff
@@ -94,12 +95,18 @@ filter_for_release <- function(dataset, selected_dataset, selected_cohort,
...
   release_cols <- intersect(release_cols, colnames(dataset))
   
-  # filter the data by cohort and release columns
+  # filter the data by cohort, release columns, remove retracted patients
   release_dat <- dataset %>%
     filter(cohort_internal==selected_cohort) %>%
     filter(!record_id %in% retracted_patient) %>%
     select(all_of(release_cols))
   
+  # remove retracted sample
```
rxu17, line 104: I think it might be best to have some unit tests for this function since it does the retraction
EDIT: Based on coordination meeting discussion, this has to wait due to deadline tmrw 12/11

scripts/release/update_retraction_table.py
```diff
@@ -1,208 +1,243 @@
...
-        help="dry run flag"
-    )
-    
+    """Main function to update the retraction table for a specific cohort."""
+    parser = argparse.ArgumentParser(description='Update retraction for release table on Synapse for BPC')
+    parser.add_argument("-c", "--cohort", help="Cohort to release (e.g., NSCLC, CRC, BrCa, BLADDER)", required=True)
+    parser.add_argument("-s", "--synapse_config", default=synapseclient.client.CONFIG_FILE, help="Path to Synapse credentials file")
+    parser.add_argument("-m", "--message", default="", help="Version comment for the table update")
+    parser.add_argument("-d", "--dry_run", action="store_true", help="Flag for dry run (no updates will be made)")
+
```
rxu17, line 130: 1. Let's add an additional argument called `--production` and create a staging version of the retraction table [here](https://www.synapse.org/Synapse:syn63600793/tables/). 
2. Then convert `RETRACTION_TABLE_ID` to like a dictionary or something to hold the two sets of synapse ids. 
3. Then add logic here to use the correct synapse id depending on the `production` flag. 

This way we can run things on staging first to see if the retractions are as expected/test the synapse upload. Thoughts?

rxu17 (approved):
LGTM! Just a comment about including staging for the retraction table update.

I think eventually we want to support staging versions of the folders/tables used in the `create_release_files.R` as well but that would require us to support the `input` and `output` and would be more time than we can do this time.

PR #168, opened 2025-02-13
[GEN-1718] Update config.json

Chelsea-Na:
Updating config to exclude all Tier1A variables for replacement to process RENAL 1.1-consortium Round 2

### DIFF
--- a/scripts/table_updates/config.json
+++ b/scripts/table_updates/config.json
@@ -29,18 +29,7 @@
     "main_genie_data_release_files": "syn16804261",
     "main_genie_sample_mapping_table": "syn7434273",
     "patient_tier1a_column_list_to_be_replaced": [
-        "naaccr_ethnicity_code",
-        "naaccr_race_code_primary",
-        "naaccr_race_code_secondary",
-        "naaccr_race_code_tertiary",
-        "naaccr_sex_code",
-        "birth_year"
     ],
     "sample_tier1a_column_list_to_be_replaced": [
-        "cpt_sample_type",
-        "cpt_seq_date",
-        "cpt_oncotree_code",
-        "cpt_seq_assay_id",
-        "age_at_seq_report"
     ]
 }


thomasyu888 (commented):
@rxu17 any way we can do a staging run before merging to make sure the code works?

Or given our bandwidth, we can also just merge and hope for the best. If there's an error, a ticket can be filed for our backlog

The good news is that this is config based, so it's a bit inconvenient to toggle it back and forth, but we can. We just have to keep track of whether we replace or not and for which cohort.

danlu1:
As what we've discussed in the [slack](https://sagebionetworks.slack.com/archives/C0350GLJ7M4/p1739250977920099?thread_ts=1738621707.971939&cid=C0350GLJ7M4), we can't bypass tier1a replacement cuz the [code](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/develop/scripts/table_updates/update_data_table.py#L570-L571) is inherited from an earlier version that subsets main genie columns from merged dataset and renames these columns with BPC column names. There are several ways to achieve this: 1. we can add a parameter to skip replacement(not calling the `custom_fix_for_tier1a_variable` [function](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/d0175a9c40932529d65e6f6990318c69d6ca5d75/scripts/table_updates/update_data_table.py#L763C9-L763C39)); 2. we can add a check on the missingness of the main genie column section in config then decide whether to overwrite or not. If we decided on what route to go, we can do staging run to test the changes out. @thomasyu888 

Apologies, but the code isn't very flexible because the original design assumes that tier1a will always be replaced.

rxu17:
I second what Dan said. The original code prior to this cohort/tier1a refactor was always going to be replacing `cpt_seq_date` and `cpt_sample_type`.  I think I'd prefer option 1 (from Dan) - adding a nextflow parameter so we don't have to do an extra step of creating a PR and modifying the config file - giving us flexibility with handling no replacement for old cohorts and replacement for new ones if needed. 

Note that for option 1, we'd also have to skip the updating version for tier1a replacement mapping tables code too right after the custom_fix_for_tier1a_variable [function](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/d0175a9c40932529d65e6f6990318c69d6ca5d75/scripts/table_updates/update_data_table.py#L763C9-L763C39)

danlu1:
@Chelsea-Na I'd like to close this PR since we will add toggle to skip tier1a replacement in another PR.

PR #170, opened 2025-02-21
[GEN-1744] add tier1a replacement param

danlu1:
Problem:

1. Currently, there is no option to disable Tier1a replacement.
2. The mapping table updates are not cohort-specific.

Solution:

1. Introduce parameters to toggle Tier1a patient and/or sample variable replacement on or off.
2. Modify the mapping table to incorporate the Tier1a replacement parameter and enable cohort-specific table queries.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -142,7 +142,9 @@ workflow BPC_PIPELINE {
         merge_and_uncode_rca_uploads.out, 
         ch_cohort, 
         ch_comment, 
-        params.production
+        params.production,
+        params.replace_patient_tier1a,
+        params.replace_sample_tier1a
     )
 
     update_date_tracking_table(
--- a/main.nf
+++ b/main.nf
@@ -26,6 +26,8 @@ params.schema_ignore_params = ""
 params.help = false
 params.step = "update_potential_phi_fields_table"
 params.use_grs = false
+params.replace_patient_tier1a = false
+params.replace_sample_tier1a = false
 
 /*
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -110,7 +112,9 @@ workflow BPC_PIPELINE {
         "default", 
         ch_cohort, 
         ch_comment, 
-        params.production
+        params.production,
+        params.replace_patient_tier1a,
+        params.replace_sample_tier1a
     )
    } else if (params.step == "genie_bpc_pipeline"){
     update_potential_phi_fields_table(ch_comment, params.production)
--- a/modules/update_data_table.nf
+++ b/modules/update_data_table.nf
@@ -12,6 +12,8 @@ process update_data_table {
    val cohort
    val comment
    val production
+   val replace_patient_tier1a
+   val replace_sample_tier1a
 
    output:
    stdout
@@ -20,12 +22,12 @@ process update_data_table {
    if (production) {
       """
       cd /root/scripts/
-      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -pd
+      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -pd -rp $replace_patient_tier1a -rs $replace_sample_tier1a
       """
    } else {
       """
       cd /root/scripts/
-      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary
+      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -rp $replace_patient_tier1a -rs $replace_sample_tier1a
       """
    }
 }
--- a/scripts/table_updates/config.json
+++ b/scripts/table_updates/config.json
@@ -37,9 +37,6 @@
     ],
     "sample_tier1a_column_list_to_be_replaced": [
         "cpt_sample_type",
-        "cpt_seq_date",
-        "cpt_oncotree_code",
-        "cpt_seq_assay_id",
-        "age_at_seq_report"
+        "cpt_seq_date"
     ]
 }
--- a/scripts/table_updates/config.json
+++ b/scripts/table_updates/config.json
@@ -33,8 +33,7 @@
         "naaccr_race_code_primary",
         "naaccr_race_code_secondary",
         "naaccr_race_code_tertiary",
-        "naaccr_sex_code",
-        "birth_year"
+        "naaccr_sex_code"
     ],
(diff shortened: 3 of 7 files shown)


scripts/table_updates/config.json
rxu17, line 1: I think the changes to `main.nf`, `nextflow_schema.json` and `update_data_table.nf` may be missing?

rxu17, line 1: I wonder if you can also test this on an older main genie release (if you haven't already, maybe at least several consortium releases old), omit the detailed naaccr columns: (e.g: `naaccr_race_code_primary`...etc) from the list to be replaced in patient and see how the script does. I don't think we'll be making fixes for any issues in this PR but good to know for future reference ...

rxu17 (commented):
LGTM! Just a couple of comments

scripts/table_updates/config.json
rxu17: Btw, the above doesn't have to be done before this PR gets merged just a thought

danlu1: ahh yes. Let me add them

danlu1: I prefer to test it in another ticket due to the time constrains.

scripts/table_updates/update_data_table.py
```diff
@@ -619,70 +623,129 @@ def custom_fix_for_tier1a_variable(
...
         column_mapping_table,
-        bpc_column_list=config["sample_tier1a_column_list_to_be_replaced"],
+        bpc_column_list=["cpt_seq_date"],
         config=config,
         logger=logger,
         cohort=cohort,
+        comment=comment
     )
+    # reformat cpt_seq_date column
+    cpt_seq_dat["cpt_seq_date"] = cpt_seq_dat["cpt_seq_date"].map(utilities.float_to_int)
```
thomasyu888, line 741: Why is this needed? also there are NULL values in center uploaded input files for SEQ_DATE, does this account for that?

danlu1: This is what implemented in the original code and it's also applied to birth_year. It won't change the NULL values since it has a [is_float](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/b30af62ff477306f58975db3d53501f35674ba86/scripts/table_updates/utilities.py#L49C1-L63C1) check in order to proceed.

thomasyu888: No problem.  haha. Thanks!

modules/update_data_table.nf
```diff
@@ -20,12 +22,12 @@ process update_data_table {
    if (production) {
       """
       cd /root/scripts/
-      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -pd
+      python update_data_table.py -p /root/scripts/config.json -c $cohort -m "$comment" primary -pd -rp $replace_patient_tier1a -rs $replace_sample_tier1a
```
rxu17, line 25: Does this work if `-rp` and `-rs` are `store_true` type of arguments in update_data_table.py? I remember from my experience it won't accept an argument (will error), and so I had to do a lot of[ if/else logic here](https://github.com/Sage-Bionetworks/GENIE-Sponsored-Projects/blob/develop/main.nf#L21-L50) to use the `--use-grs` flag which is `store_true`. Because propagating true / false values from nextflow -> python doesn't propagate correctly

main.nf
```diff
@@ -110,7 +112,9 @@ workflow BPC_PIPELINE {
         "default", 
         ch_cohort, 
         ch_comment, 
-        params.production
+        params.production,
```
rxu17, line 115: I think these are missing for the call to `update_data_table` under `params.step == "genie_bpc_pipeline"` section

modules/update_data_table.nf
danlu1: Yes, I figured and tweaked it to be a regular kwarg on my local last week. Let me push my change.

scripts/table_updates/update_data_table.py
```diff
@@ -716,6 +782,20 @@ def main():
         action="store_true",
         help="Save output to production folder",
     )
+    parser.add_argument(
+        "-rp",
+        "--replace_patient_tier1a",
+        type = lambda x: x.lower()== 'true',
```
rxu17, line 788: Neat!

scripts/table_updates/update_data_table.py
```diff
@@ -706,9 +706,12 @@ def custom_fix_for_cpt_seq_data(
         config: dict,
         cohort: str = "",
         comment: str = "",
+        replace_patient_tier1a: bool = False, 
+        replace_sample_tier1a: bool = False, 
     ) -> None:
-    # unlist form column in master table
-    master_table["form"] = master_table["form"].apply(lambda x: ", ".join(x))
+    # unlist form column in master table only when not replacing tier1a variables
```
rxu17, line 712: Why is this needed? Can you put this in the comment description

scripts/table_updates/tests/test_utilities.py
```diff
@@ -313,21 +349,30 @@ def get__update_tier1a_data_replacement_mapping_table_test_cases():
...
     with patch.object(syn, "get", return_value=table_schema) as patch_get, patch.object(
         syn, "tableQuery"
-    ) as patch_table_query, patch.object(syn, "store") as patch_store, patch.object(syn, "delete") as patch_delete:
+    ) as patch_table_query, patch.object(syn, "store") as patch_store, patch.object(syn, "delete") as patch_delete, patch('utilities.update_version') as update_version:
         patch_table_query.return_value = MagicMock(etag="test_etag")
+        logger = MagicMock(spec=logging.Logger)
+
+        comment = "test comment"
         # Call the function
         update_tier1a_data_replacement_mapping_table(
```
rxu17, line 364: I think tests (multiple tests if not already present) for this function `update_tier1a_data_replacement_mapping_table` needs to be updated to include testing for the new parameters: `bpc_column_list`, `main_genie_column_list`.

scripts/table_updates/update_data_table.py
danlu1: Because the form column [has already been reformatted](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/b30af62ff477306f58975db3d53501f35674ba86/scripts/table_updates/update_data_table.py#L643) if we turn on replace_patient_tier1a or replace_sample_tier1a. And we do need to unlist this column if none of the tier1a toggle is on

scripts/table_updates/tests/test_utilities.py
danlu1: Sure. I will tweak this.

rxu17 (approved):
LGTM!

PR #172, opened 2025-02-25
[GEN-1767] Temp remove unlist form column code for prod table

rxu17:
**Purpose:** This comments out the code that unlists the `form` column in the fileview table (assumes that `form` column is list type) because our production table's `form` column is `string` not list so having that line throws an error. Note that this is a temp fix until investigation concludes.

See [investigation details](https://sagebionetworks.jira.com/browse/GEN-1767?focusedCommentId=238862) for more context on this issue.

**Testing:** 
- Tested on production tables up to that point and made sure the synapse_id was retrieved correctly. Cannot run the rest of the production pipeline. Not going to touch the staging pipeline (still investigating)


danlu1 (approved):
LGTM

PR #174, opened 2025-03-06
[GEN-1807] Update config.yaml

Chelsea-Na:
1. Updating CRC VICC target cases from 226 -> 225
2. Updating Melanoma cohort's target cases

All these updates per https://aacr.ent.box.com/s/h4wkij15geg7ekls583b7wdydrbu7z2w

### DIFF
--- a/scripts/case_selection/config.yaml
+++ b/scripts/case_selection/config.yaml
@@ -176,8 +176,8 @@ phase:
             production: 0
             adjusted: 0
           VICC:
-            production: 226
-            adjusted: 226
+            production: 225
+            adjusted: 225
             sdv: 45
             irr: 11
         date:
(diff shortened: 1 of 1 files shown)


scripts/case_selection/config.yaml
rxu17, line 1: It looks like these numbers are just used in the [outputted case count table for display](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/ee5a9afa4ae859421ddddf622074a94c454707b7/scripts/case_selection/update_case_count_table.R#L184-L200) not used elsewhere during the case_selection code?

rxu17 (approved):
LGTM!

scripts/case_selection/config.yaml
Chelsea-Na: That is correct. As a note we also use these numbers in genie-bpc-quac to make sure sites are uploading the correct case counts

PR #177, opened 2025-04-03
[GEN-1874] convert tier1a replacement mapping table to long format

danlu1:
# **Problem:**

1. The current structure of the cancer_panel_test_tier1a_replacement_mapping_table and patient_characteristics_tier1a_replacement_mapping_table poses significant challenges for quality control and comparison tasks. Specifically, it requires constant schema updates whenever new columns are introduced for replacement—especially when using non-_DETAILED columns—and leads to excessive NULL values when certain columns are not applicable to specific consortium releases. To address these limitations, we are restructuring the tables to create a more flexible and scalable design that simplifies QC processes and reduces unnecessary data sparsity.

2. site column need to be added to the table

# **Solution:**
1. add function to extract site name from the id column
2. add function to convert the mapping table from wide to long format. 

# **Testing:**
Relevant test functions have been added or updated.

### DIFF
--- a/scripts/table_updates/tests/test_utilities.py
+++ b/scripts/table_updates/tests/test_utilities.py
@@ -519,53 +519,95 @@ def test_update_tier1a_data_replacement_mapping_table(
 
 
 def test_extract_site_name_from_sample_id_single_str():
-    cpt_genie_sample_id = "GENIE-1"
-    assert extract_site_name_from_sample_id(cpt_genie_sample_id) == "1"
+    cpt_genie_sample_id = "GENIE-SAGE-1"
+    assert extract_site_name_from_sample_id(cpt_genie_sample_id) == "SAGE"
 
 
-def test_convert_mapping_table_to_long():
-    # Sample input DataFrame
-    df = pd.DataFrame(
-        {
-            "cohort": ["A", "A"],
-            "cpt_sample_id": ["cohortA-site1-sample01", "cohortA-site1-sample02"],
-            "bpc_col1": [10, 20],
-            "bpc_col2": [30, 40],
-            "genie_col1": [50, 60],
-            "genie_col2": [70, 80],
-        }
-    )
-
+@pytest.mark.parametrize(
+    "input_df,expected_df",
+    [
+        (
+            pd.DataFrame(
+                {
+                    "cohort": ["A", "A"],
+                    "cpt_sample_id": [
+                        "cohortA-site1-sample01",
+                        "cohortA-site1-sample02",
+                    ],
+                    "bpc_col1": [10, 20],
+                    "bpc_col2": [30, 40],
+                    "genie_col1": [50, 60],
+                    "genie_col2": [70, 80],
+                }
+            ),
+            pd.DataFrame(
+                {
+                    "cohort": ["A", "A", "A", "A"],
+                    "cpt_sample_id": [
+                        "cohortA-site1-sample01",
+                        "cohortA-site1-sample02",
+                        "cohortA-site1-sample01",
+                        "cohortA-site1-sample02",
+                    ],
+                    "bpc_field": ["bpc_col1", "bpc_col1", "bpc_col2", "bpc_col2"],
+                    "bpc_field_value": [10, 20, 30, 40],
+                    "main_genie_field": [
+                        "genie_col1",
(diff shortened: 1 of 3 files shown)


scripts/table_updates/tests/test_utilities.py
```diff
@@ -379,6 +508,66 @@ def test_update_tier1a_data_replacement_mapping_table(syn, table_schema, test_ca
...
+            "Updating version for tier1a data replacement mapping table"
+        )
+        update_version.assert_called_with(
+            syn,
+            table_schema.id,
+            f"{comment}_mainGENIE_{config['main_genie_release_version']}",
+        )
+
+
+def test_extract_site_name_from_sample_id_single_str():
```
rxu17, line 521: nit: Might be good to use actual sample_id example here like "GENIE-SAGE-1"

scripts/table_updates/utilities.py
```diff
@@ -212,8 +212,71 @@ def remove_backslash(df: pandas.DataFrame, cols: List[str]) -> pandas.DataFrame:
...
+    Args:
+        cpt_genie_sample_id (str): sample ID
+
+    Returns:
+        str: Site name
+    """
+    return sample_id.split("-")[1]
+
+
+def convert_mapping_table_to_long(
```
rxu17: nit: for consistency, function can match other tier1a function names like `convert_tier1a_data_replacement_mapping_table_to_long()`

scripts/table_updates/utilities.py
```diff
@@ -212,8 +212,71 @@ def remove_backslash(df: pandas.DataFrame, cols: List[str]) -> pandas.DataFrame:
...
+    df: pandas.DataFrame,
+    id_vars: List[str],
+    bpc_column_list: List[str],
+    main_genie_column_list: List[str],
+) -> pandas.DataFrame:
+    """Convert a the dataframe from wide to long format
```
rxu17: Nit: could use more details here in the description

scripts/table_updates/tests/test_utilities.py
rxu17, line 1: Wondering if we should have test cases with NAs in them and make sure behavior with reshaping is as expected for those? 

Also wondering if we should have test cases with input data of different data types, I'm not sure from your experience if the BPC mapping data could possibly have different data types. I see that for example the [old mapping table](https://www.synapse.org/Synapse:syn64331055/tables/) are all string columns, was this the case for the input data too? This may not be necessary... but just thinking out loud for testing unexpected reshape behavior esp. with different data types.

rxu17 (approved):
LGTM! Just some minor comments

scripts/table_updates/tests/test_utilities.py
danlu1: > Wondering if we should have test cases with NAs in them and make sure behavior with reshaping is as expected for those?

I can add the test cases. Though I believe melt function can handle NAs. 

> Also wondering if we should have test cases with input data of different data types, I'm not sure from your experience if the BPC mapping data could possibly have different data types. I see that for example the [old mapping table](https://www.synapse.org/Synapse:syn64331055/tables/) are all string columns, was this the case for the input data too? This may not be necessary... but just thinking out loud for testing unexpected reshape behavior esp. with different data types.

I believe Synapse will store column values as strings regardless of the original data format. Here is an example to save integer data to a Synapse [table](https://www.synapse.org/Synapse:syn65916870/tables/) with all string columns:
 ```
import pandas as pd
import synapseclient
from synapseclient import Table

syn = synapseclient.Synapse()
syn.login()

table_schema = syn.get('syn65916870')

df  = pd.DataFrame({
    "col1": [1, 2, 3, 4, 5],
    "col2": [10, 20, 30, 40, 50]
})

syn.store(Table(table_schema, df))
```
gives: 

Uploading: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 194B/s, table.csv]
Applying an update to table: 'syn65916870' ... [--------------------]0.00%   0/100       
Applying an update to table: 'syn65916870' ... [####################]100.00%   100/100   Done...

danlu1: Confirmed in the test cases that melt function can deal with NAs.

PR #180, opened 2025-04-19
[GEN-1943]Address nan in sample_id column when extracting site name

danlu1:
# **Problem:**

AttributeError: 'float' object has no attribute 'split' in `extract_site_name_from_id` function

# **Solution:**
convert float('nan') to empty string and extract site name when id value is not empty string

# **Testing:**
unit test is added and integration test is done.

### DIFF
--- a/scripts/table_updates/utilities.py
+++ b/scripts/table_updates/utilities.py
@@ -212,17 +212,23 @@ def remove_backslash(df: pandas.DataFrame, cols: List[str]) -> pandas.DataFrame:
         raise ValueError("Invalid column list. Not all columns are in the dataframe.")
 
 
-def extract_site_name_from_sample_id(sample_id: str) -> str:
-    """Extract site name from CPT Genie Sample ID
-       Site name is the second substring of the sample ID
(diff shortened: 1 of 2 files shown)


scripts/table_updates/utilities.py
```diff
@@ -212,17 +212,23 @@ def remove_backslash(df: pandas.DataFrame, cols: List[str]) -> pandas.DataFrame:
...
 
     Returns:
         str: Site name
     """
-    return sample_id.split("-")[1]
+    # convert nan to empty string
+    id = "" if pandas.isna(id) else str(id)
+    if id:
+        # Split the ID by "-" and return the second substring if the ID is not empty
+        return id.split("-")[1]
```
thomasyu888, line 229: Are all the values GENIE-site in bpc?

thomasyu888 (approved):
🔥 LGTM! Thanks for the quick fix

scripts/table_updates/utilities.py
danlu1: Yes, unless it's NA.

PR #181, opened 2025-04-29
[GEN-1962] add yesno mapping to the mapping when grs is not used

xindiguo:
# **Problem:**

When the grs toggle is set to no, the yesno variables are not included in the mappings object, which causes those variables do not map to any text labels.

See detailed comments in https://sagebionetworks.jira.com/browse/GEN-1961

# **Solution:**

Added the list of yesno variables to the mappings object with 0=No, 1=Yes as mapping


thomasyu888 (approved):
LGTM! did you want to add any tests for this in the test folder? I know it's urgent but it may save us in the long run.

I'll defer to you.

xindiguo:
> LGTM! did you want to add any tests for this in the test folder? I know it's urgent but it may save us in the long run.
> 
> I'll defer to you.

Thanks! I have added the test specifically for the `use_grs=FALSE` case

PR #182, opened 2025-04-29
[GEN-1836] Replace AGE_AT_SEQ_REPORT with AGE_AT_SEQ_REPORT_DAYS

xindiguo:
# **Problem:**

The [perform_case_selection.R](https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/develop/scripts/case_selection/perform_case_selection.R)  incorrectly using AGE_AT_SEQ_REPORT (in years) instead of AGE_AT_SEQ_REPORT_DAYS (in days), resulting in inaccurate case selection.

# **Solution:**

Replaced all four instances of AGE_AT_SEQ_REPORT with AGE_AT_SEQ_REPORT_DAYS in perform_case_selection.R to ensure correct unit usage.

### DIFF
--- a/scripts/case_selection/perform_case_selection.R
+++ b/scripts/case_selection/perform_case_selection.R
@@ -180,7 +180,7 @@ get_eligibility_data <- function(synid_clinical, site) {
            SAMPLE_ID, 
            ONCOTREE_CODE, 
            SEQ_DATE, 
-           AGE_AT_SEQ_REPORT,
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM! 

Due to urgency - I know these don't have tests so it's fine, but we may want some sort of safe guard against these changing values.

PR #183, opened 2025-05-01
[GEN-1837] Add age max filter

xindiguo:
# **Problem:**
We would like to restrict the age upper limit during case selection process. The age_max flag is already in the current configuration file. Global setting is Inf. 

# **Solution:**

- Add related helper functions to get age_max in years and days
- Add the age_max filter to the create_eligibility_matrix function using age_max in days
- update the rmarkdown using age_max in years

# **Testing:**
Currently there is no testing. Also the filter logic needs to be refined


thomasyu888 (approved):
LGTM

PR #184, opened 2025-05-07
[GEN-1986] Update config.yaml

Chelsea-Na:
Update config file per Alyssa’s email on Apr 30, 2025 titled “BPC Esophagogastric Case Lists” using the numbers [here](https://aacr.ent.box.com/s/a5jf5zx6l1g0ddv7pvjnxm3vauzi9szk)

- Update pressure test cases per site
- Update seq date range
- Update max age to 85 years of age. This means we are selecting patients that are less than 31025 (365*85) days at age of sequencing.

### DIFF
--- a/scripts/case_selection/config.yaml
+++ b/scripts/case_selection/config.yaml
@@ -560,10 +560,12 @@ phase:
           DFCI:
             production: 315
             adjusted: 315
+            pressure: 15
             irr: 0
           MSK:
             production: 515
             adjusted: 515
+            pressure: 15
             irr: 0
           PROV:
             production: 100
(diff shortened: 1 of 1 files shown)


scripts/case_selection/config.yaml
```diff
@@ -572,30 +574,36 @@ phase:
...
             irr: 0
           WAKE:
             production: 92
             adjusted: 92
             irr: 0
         date: 
           seq_min: Apr-2012
-          seq_max: Feb-2023
+          seq_max: Oct-2022
+        age_max: 31025
```
thomasyu888, line 606: Will you check this after merge? Im unsure if we do age_max * 365

thomasyu888 (approved):
🔥 LGTM!

PR #185, opened 2025-05-07
[GEN-1986] Update config.yaml

Chelsea-Na:
Updating age_max unit for the esophago cohort from days to years.

### DIFF
--- a/scripts/case_selection/config.yaml
+++ b/scripts/case_selection/config.yaml
@@ -603,7 +603,7 @@ phase:
         date: 
           seq_min: Apr-2012
           seq_max: Oct-2022
-        age_max: 31025
+        age_max: 85
         oncotree:
           root: EGC
           allowed_codes: 


scripts/case_selection/config.yaml
```diff
@@ -603,7 +603,7 @@ phase:
         date: 
           seq_min: Apr-2012
           seq_max: Oct-2022
-        age_max: 31025
+        age_max: 85
```
thomasyu888, line 606: Years now? Also.   Does this need to be 86?

Chelsea-Na: I realized that there is this in scripts/case_selection/shared_fxns.R  that converts to age_max in days. 

```
get_age_max_days <- function(config, phase, cohort) {
  return(get_age_max(config, phase, cohort)*365)
}
```

I also just wanted to test 85 years of age to see if/how it works. AACR still has not confirmed how we want to interpret 85 years in days we will use.

thomasyu888 (approved):

PR #186, opened 2025-05-09
[GEN-1786] update randomization method for SDV cases

danlu1:
# **Problem:**

We need to update the script where each patient has a 20% chance of being marked as SDV instead of making sure X number of production cases meet a certain target of SDV cases.

# **Solution:**

We simulate Bernoulli trails to flag sdv cases. 

# **Testing:**

(wip)


thomasyu888 (approved):
🔥 LGTM. I'll defer to @Chelsea-Na for final testing of the solution and cross checking the esophago cases selected

PR #187, opened 2025-05-10
[GEN-1783] Update perform_case_selection.R age max logic

Chelsea-Na:
# **Problem:**

In order for us to set an age max of 85 years of age (e.g. 85.999 years of age patients is eligible for in cohort select), we need to update the age max logic to be < and not <=. We will also have to make it so the age_max is +1 of what the colloquial term for age_max is. 

# **Solution:**

Update this section: `AGE_AT_SEQ_REPORT_DAYS <= age_max` to reflect `AGE_AT_SEQ_REPORT_DAYS < age_max` 

# **Testing:**

No testing done.

### DIFF
--- a/scripts/case_selection/config.yaml
+++ b/scripts/case_selection/config.yaml
@@ -603,7 +603,7 @@ phase:
         date: 
           seq_min: Apr-2012
           seq_max: Oct-2022
-        age_max: 85
+        age_max: 86
         oncotree:
           root: EGC
           allowed_codes: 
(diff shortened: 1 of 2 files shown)


scripts/case_selection/config.yaml
```diff
@@ -603,7 +603,7 @@ phase:
         date: 
           seq_min: Apr-2012
           seq_max: Oct-2022
-        age_max: 85
```
thomasyu888, line 606: Are there other age_max variables in the config we should fix?

thomasyu888 (approved):

scripts/case_selection/config.yaml
Chelsea-Na: No, the other cohorts already have had their cases selected. Esophago cohort is the last one.

PR #188, opened 2025-05-12
[GEN-1786] update randomization method

danlu1:
# **Problem:**

1. The perform_case_selection.R could not be sourced without arguments. 
2. No unit test for create_selection_matrix function
3. Docker image is not built automatically for case_selection step

# **Solution:**

1. Refactor the variable setting and check sections within perform_case_selection.R as a main function which will be called only when the script is running in a interactive mode.
2. Add unit test for create_selection_matrix function
3. Edit workflow file to allow building docker for case_selection module.

# **Testing:**
1. Integration test: `Rscript perform_case_selection.R -p 2 -c ESOPHAGO -s MSK -r 17.6-consortium` ran through successfully.
2. Unit tests for create_selection_matrix passed.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        module: ["references", "table_updates", "uploads"] # Define the modules you want to loop through for builds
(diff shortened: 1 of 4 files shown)


thomasyu888 (approved):
🔥 LGTM!

thomasyu888 (approved):
🔥 LGTM!  I'll defer to @Chelsea-Na for a final review of the outputs

PR #189, opened 2025-05-14
[GEN-1996] update table schema

thomasyu888:
# **Problem:**
The ovarian cohort had a new data dictionary which requires us to run update_table_schema.py code to make sure new columns were added.

While running this code, a failure occurred
```
Traceback (most recent call last):
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 203, in <module>
    main()
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 200, in main
    update_table_schema(syn,logger,dry_run)
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 167, in update_table_schema
    _update_table_schema(syn, form, curated_data_element, logger, dry_run)
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 82, in _update_table_schema
    non_check_new_cols = list(non_check_to_add_df.apply(lambda x: create_synapse_column(x['variable'],x['synColType'],x['synColSize']),axis=1))
  File "/Users/tyu/anaconda3/envs/synapse/lib/python3.10/site-packages/pandas/core/frame.py", line 10034, in apply
    return op.apply().__finalize__(self, method="apply")
  File "/Users/tyu/anaconda3/envs/synapse/lib/python3.10/site-packages/pandas/core/apply.py", line 837, in apply
    return self.apply_standard()
  File "/Users/tyu/anaconda3/envs/synapse/lib/python3.10/site-packages/pandas/core/apply.py", line 965, in apply_standard
    results, res_index = self.apply_series_generator()
  File "/Users/tyu/anaconda3/envs/synapse/lib/python3.10/site-packages/pandas/core/apply.py", line 981, in apply_series_generator
    results[i] = self.func(v, *self.args, **self.kwargs)
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 82, in <lambda>
    non_check_new_cols = list(non_check_to_add_df.apply(lambda x: create_synapse_column(x['variable'],x['synColType'],x['synColSize']),axis=1))
  File "/Users/tyu/sage/genie-bpc-pipeline/scripts/table_updates/update_table_schema.py", line 32, in create_synapse_column
    maximumSize=int(max_size))
ValueError: cannot convert float NaN to integer
```

# **Solution:**
hacky workaround to auto fill in string of length 100.  This could fail for any column that has longer string length than 100 for a particular row

# **Testing:**
Ran the code - no tests.

### DIFF
--- a/scripts/table_updates/update_table_schema.py
+++ b/scripts/table_updates/update_table_schema.py
@@ -27,7 +27,7 @@ def create_synapse_column(name, col_type, max_size):
         new_column = Column(name=name,
                             columnType=col_type)
     else:
-        max_size = 100 if pandas.isna(max_size) else int(max_size)
+        max_size = 500 if pandas.isna(max_size) else int(max_size)
         new_column = Column(name=name,
                             columnType=col_type, 
                             maximumSize=max_size)
--- a/scripts/table_updates/update_table_schema.py
+++ b/scripts/table_updates/update_table_schema.py
@@ -27,9 +27,10 @@ def create_synapse_column(name, col_type, max_size):
         new_column = Column(name=name,
                             columnType=col_type)
     else:
+        max_size = 100 if pandas.isna(max_size) else int(max_size)
         new_column = Column(name=name,
                             columnType=col_type, 
-                            maximumSize=int(max_size))
+                            maximumSize=max_size)
     return new_column
 
 def _expand_checkbox_vars(row):
--- a/scripts/table_updates/README.md
+++ b/scripts/table_updates/README.md
@@ -22,9 +22,6 @@ Please make sure you have the [.synapseConfig file](https://help.synapse.org/doc
 Make sure you have Python 3.8 and conda installed.
 
 ### Install the required packages
-```
-pip install 'synapseclient[pandas] == 2.7.2'
-```
 
 > [!NOTE]  
 > Due to this tool using an older version of the python client, until there is bandwidth to update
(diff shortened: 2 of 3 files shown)


scripts/table_updates/update_table_schema.py
```diff
@@ -27,9 +27,10 @@ def create_synapse_column(name, col_type, max_size):
         new_column = Column(name=name,
                             columnType=col_type)
     else:
+        max_size = 100 if pandas.isna(max_size) else int(max_size)
```
danlu1: Can we go extreme and set this as 500? I run into several times that I have to modify the column size from 100~ to a bigger number.

danlu1 (approved):
LGTM. Just one nit.

scripts/table_updates/update_table_schema.py
thomasyu888:
```suggestion
        max_size = 500 if pandas.isna(max_size) else int(max_size)
```

thomasyu888: Thanks @danlu1 , my only concern with updating this to a larger number is that we can potentially run into table size limits. Since there are so many string columns, the Tables string length is a pre-defined length.  So even if the values within the Table don't take up that length, Synapse will not allow for the sum of the string column maximum size constraint to be greater than the allowable byte size per row.

I did just push the update though

danlu1: Interesting. I didn't know there is a limit on row though.

PR #191, opened 2025-06-04
[GEN-2039] Update config.json

Chelsea-Na:
For the prostate 1.0-public release, we will have to use the 10.3-consortium release to replace the tier 1a values. See data lock column here: syn20945902


rxu17 (approved):
LGTM!

PR #195, opened 2025-06-19
[GEN-1937] Update config.json

Chelsea-Na:
Updating to use 13.6-consortium to replace the Tier 1A data for the CRC Phase 1 and 2 cohort per the cohort's data lock date.


rxu17 (approved):
LGTM!

PR #196, opened 2025-07-11
[GEN-2089] Update config.json

Chelsea-Na:
Updating for the BrCa cohort. See ticket https://sagebionetworks.jira.com/browse/GEN-2089


rxu17 (approved):
LGTM! I would double check that the dockerhub image builds without error after this merges: https://hub.docker.com/r/sagebionetworks/genie-bpc-pipeline-table-updates

PR #200, opened 2025-08-13
[GEN-1773] Implement `params.entry` due to Nextflow's workflow entry option deprecation

rxu17:
# **Problem:**
There will be a breaking change for when Seqera Platform upgrades to Nextflow >= 24.10.0. The [`entry` option](https://www.nextflow.io/docs/latest/reference/cli.html#cli-reference) will become deprecated.

Example of our case selection workflow that uses this to-be deprecated`entry` when we specify by workflow entry name in the nextflow UI
```
nextflow run 'https://github.com/Sage-Bionetworks/genie-bpc-pipeline'
		 -name ESOPHAGO_VICC_2_case_selection
		 -r develop
		 -profile aws_prod
		 -entry case_selection
```

# **Solution:**
See this [PR in nf-synapse](https://github.com/Sage-Bionetworks-Workflows/nf-synapse/pull/11) for the original method that was copy and pasted here. 

Implemented `params.entry` instead of using nextflow's to-be deprecated entry option. Created profiles for each of our bpc workflows which set the `params.entry` accordingly.

# **Testing:**
- Added parameter validation
- Launched on Nextflow (removed entry workflow spec and launched with `bpc_pipeline` profile)
- Launched on Nextflow (removed entry workflow spec and launched with `case_selection` profile)

Testing report here: 
https://sagebionetworks.jira.com/browse/GEN-1773?focusedCommentId=263458

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,11 +217,21 @@ This workflow takes care of transferring files to and from Synapse. Hence, it re
 1. Generate a personal access token (PAT) on Synapse using [this dashboard](https://www.synapse.org/#!PersonalAccessTokens:). Make sure to enable the `view`, `download`, and `modify` scopes since this workflow both downloads and uploads to Synapse.
 2. Create a secret called `SYNAPSE_AUTH_TOKEN` containing a Synapse personal access token using the [Nextflow CLI](https://nextflow.io/docs/latest/secrets.html)
 
-### Commands
+### Running the pipeline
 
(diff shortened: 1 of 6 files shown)


thomasyu888 (commented):
🔥 Thanks for doing this.  I'll defer to @danlu1 for final review of this!

danlu1 (approved):
LGTM. Thanks for working on this.

PR #204, opened 2025-09-08
[GEN-2025]fix ValueError: Cannot index with multidimensional key in data catalog update script

danlu1:
# **Problem:**
The issue: "ValueError: Cannot index with multidimensional key" was found earlier but being forgotten to be added to the original PR. 
Screenshot 2025-09-08 at 11 54 19 AM

# **Solution:**
Add the fix. 

# **Testing:**

The issue is solved by including the fix. 
Screenshot 2025-09-08 at 11 54 11 AM


thomasyu888 (approved):
🔥 What was this originally printing out?

danlu1:
@thomasyu888 should be "2025-09-08 18:45:42 INFO     Updated both columns size and number: X " to indicate the number of variable that being updated size and key number.

PR #207, opened 2025-11-05
[GEN-2043] Refactor timezone conversion in update_cbio_mapping.py

xindiguo:
Updated timezone handling for modified timestamps using pytz.

# **Problem:**
Refactor timezone conversion due to synutils cannot be used.

# **Solution:**
Update to use pytz instead

### DIFF
--- a/scripts/references/update_cbio_mapping.py
+++ b/scripts/references/update_cbio_mapping.py
@@ -1,10 +1,10 @@
 import argparse
 from datetime import datetime
-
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
This looks good, I would take a look at the copilot suggestion (pre-approving), and consider splitting this into a helper function and add a unit test so that we know the expected behavior. No need to add to a CI to it or add tests for the entire thing

PR #214, opened 2026-01-27
[GEN-2473] Refactor: migrate to synapseclient.models.Table API

xindiguo:
# **Problem:**
The script was using the deprecated/legacy Synapse Table API methods:
- syn.tableQuery("SELECT...").asDataFrame() for querying tables
- syn.store(Table(schema, dataframe, etag=...)) for updating rows
- Required manual etag management to handle concurrent updates
- Legacy Table import from synapseclient conflicted with new models.Table

# **Solution:**
- Replace syn.tableQuery().asDataFrame() with Table(id=).query() method
- Replace syn.store(Table(...)) with:
  - Table.upsert_rows(values=df, primary_keys=["variable"]) for updates
  - Table.store_rows(values=df) for new row insertion
- Move 'from synapseclient.models import Table' after utilities import
  to properly override the legacy Table reference
- Remove manual etag handling as new API manages this internally

This ensures compatibility with the latest synapseclient library and follows the recommended patterns for Synapse Table operations.

### DIFF
--- a/scripts/table_updates/update_data_element_catalog.py
+++ b/scripts/table_updates/update_data_element_catalog.py
@@ -9,8 +9,9 @@ import re
 import pandas
 import synapseclient
 from synapseclient import Column
-from synapseclient.models import Table
 from utilities import *
+# Import Table from models AFTER utilities to override the legacy Table import
+from synapseclient.models import Table
 
 
 def set_up(args: argparse.Namespace):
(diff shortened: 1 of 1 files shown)


scripts/table_updates/update_data_element_catalog.py
```diff
@@ -268,25 +269,33 @@ def update_by_data_dictionary(args: argparse.Namespace) -> None:
...
+                primary_keys=["variable"]
             )
             saved_to_table = True
 
         if not vars_to_add_df.empty:
             vars_to_add_df = _create_new_row(vars_to_add_df, cohort)
             # add the new cohort_dd column to the table schema
-            if "%s_dd" % cohort not in results.asDataFrame().columns:
+            # Check if column exists by querying
+            existing_df = Table(id=catalog_id).query(
```
rxu17, line 290: Do you know how big the catalog_id tables are typically? I worry that the table may not be available to query so soon after it's been updated (above).

scripts/table_updates/update_data_element_catalog.py
rxu17, line 1: Given that this script belongs in a module `table_updates` with other scripts and a Dockerfile/requirements.txt file and I don't see that the other scripts have been updated to use the new synapseclient models yet. 

I would make a note of the specific synapseclient version to use to run this script specifically in the README and not to use the usual Dockerfile/requirements txt file setup as that won't work.

